PR #22, opened 2025-04-26
Add LM Studio support

pwilkin:
Added LM Studio support, separately from Ollama support so both of those can be utilized independently.


wwwillchen:
Thanks for the PR.

Instead of adding the @lmstudio/sdk dependency, I think we can just rely on the OpenAI-compatible API - I'd like to avoid dependencies when reasonable to keep the download size as manageable as possible.

We should be able to fetch the LM Studio models using the OpenAI compatible that LM studio provides: https://lmstudio.ai/docs/app/api/endpoints/openai

I see that you're using the OpenAI client to do the model call, but you can actually use: https://sdk.vercel.ai/providers/openai-compatible-providers/lmstudio.

pwilkin:
> Instead of adding the @lmstudio/sdk dependency, I think we can just rely on the OpenAI-compatible API - I'd like to avoid dependencies when reasonable to keep the download size as manageable as possible.
> We should be able to fetch the LM Studio models using the OpenAI compatible that LM studio provides: https://lmstudio.ai/docs/app/api/endpoints/openai

There's a problem with that, though - the OpenAI-compatible `/v1/models` path lists *loaded* models, not available ones. So we'd have to manually fetch from the LM Studio-specific endpoint (not that it's a big problem).

> I see that you're using the OpenAI client to do the model call, but you can actually use: https://sdk.vercel.ai/providers/openai-compatible-providers/lmstudio.

You're right :)

Okay, I'll try to remove the LMStudio SDK (it's not very well documented anyways) and just do a manual fetch from the models endpoint to get the models.

pwilkin:
Alright, I removed the references and added the manual fetch. While I was at it, I also fixed all linter errors so that npm run lint returns 0 errors :)

wwwillchen:
Let's keep this PR tightly focused to LM studio only changes - I appreciate the lint fixes, but it's harder to review.

pwilkin:
> Let's keep this PR tightly focused to LM studio only changes - I appreciate the lint fixes, but it's harder to review.

Aight, will redo.

pwilkin:
Sorry, haven't had time for a few days, but I cleaned up the PR and rebased on current main branch.

package.json
wwwillchen, line 1: I think all the changes in this file should be reverted except for @ai-sdk/openai-compatible

package-lock.json
wwwillchen, line 1: see below  - revert

.github/workflows/release.yml
wwwillchen, line 1: revert

src/components/ModelPicker.tsx
```diff
@@ -63,21 +82,24 @@ export function ModelPicker({
...
-  const hasLocalModels =
-    !localModelsLoading && !localModelsError && localModels.length > 0;
+  // Determine availability of local models
+  const hasOllamaModels = !ollamaLoading && !ollamaError && ollamaModels.length > 0;
+  const hasLMStudioModels = !lmStudioLoading && !lmStudioError && lmStudioModels.length > 0;
 
   return (
-    <DropdownMenu open={open} onOpenChange={setOpen}>
+    <DropdownMenu open={open} onOpenChange={(newOpen) => {
+      console.log("DropdownMenu onOpenChange:", newOpen);
```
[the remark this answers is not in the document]
wwwillchen: remove

src/ipc/handlers/local_model_ollama_handler.ts
wwwillchen, line 1: remove all console logs - use logger if something needs to be logged

wwwillchen (changes requested):
Overall changes look good but need some code clean-ups and reverting earlier changes before we can merge this.

UX-wise, I think Ollama and LM Studio models should be grouped into one category "Local models" - they can still be in their own sub-category (e.g sub-sub-category), but I can take care of this after this PR if you want.

package.json
pwilkin: Done.

package-lock.json
pwilkin: Done.

.github/workflows/release.yml
pwilkin: Done.

src/ipc/handlers/local_model_ollama_handler.ts
pwilkin: Done.

pwilkin:
Aight, made the changes, as for the button logic, I'd leave it to you - I'm *terrible* at layouting :>

wwwillchen (approved):
Thanks! Merging

Issue #21, opened 2025-04-25
Support custom OpenAI-compatible API endpoint


wwwillchen:
In particular for local model providers like ollama and LM studio

Nolanaschwartz:
@wwwillchen Added some enhancements/fixes for someone like me who self-hosts on my local network, let me know if I can open a PR with these changes!

1. Local Network models:

![Image](https://github.com/user-attachments/assets/a8d42ad5-6dae-49c3-9c82-d16a851ea280)

2. URL Config in settings page:

![Image](https://github.com/user-attachments/assets/24977324-0091-4c72-a955-f3a8d45c45f0)

wwwillchen:
Thanks for sharing this. I'm going to work on revamping the settings page in next week's release which will support this and #54. If I don't address these knobs with the next release (v0.5.0), please feel free to open a PR. 

The main question in my mind is the UX since the settings page is already getting a bit cluttered.

wwwillchen:
Launched custom model support in [v0.5.0](https://www.dyad.sh/docs/releases/0.5.0) - docs on how to use it: https://www.dyad.sh/docs/guides/ai-models/custom-models

It's not quite as convenient as customizing LM Studio / ollama URL, but you should be able to accomplish this by configuring a new custom AI provider (and set it to any URL).

PR #107, opened 2025-05-07
Add local llm url config

Nolanaschwartz:
As mentioned in #21 this is the minimal setup I used to configure local models. Not sure if I missed anything since I am not familiar with the codebase, but feel free to use this as a jumping off point!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,4 @@ out/
 sqlite.db
 userData/
 .env.local
+.idea/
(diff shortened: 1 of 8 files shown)


wwwillchen:
Thanks! I'll take a look next week and see if I can incorporate some of these changes in next week's release.

wwwillchen:
see https://github.com/dyad-sh/dyad/issues/21#issuecomment-2881468488 - this is basically supported in dyad now

PR #387, opened 2025-06-11
feat: add Vercel integration and extended file type support in codebase.ts

Nokitomo:
This pull request introduces:

- Full Vercel integration, including UI components, routes, and handler logic.
- Support for additional file extensions in `codebase.ts` to enhance file parsing and indexing.
- Manual conflict resolution after rebasing onto latest main.
- Updates to schema, preload, and settings pages to reflect new functionality.


wwwillchen:
Thanks for opening this pull request! Really appreciate the effort here.

Because this is a large change that's affecting a lot of files, it'll take me a while to review this. To make this easier to review, could we split up these changes into two separate pull requests?

I think doing an initial pull request for additional file types would be good. This PR adds a lot of file types and some of them I'm hesitant about including (e.g. lock files and configs) because it would eat up a lot of the AI's token context and don't provide a lot of benefit. I think just adding a few specific file types that you need should be relatively straightforward. If there's a lot of different file types you need or unusual ones, I think it would be better to add a setting for this (e.g. user types in custom file extensions which are displayed in chips).

For the Vercel change, I think overall this makes sense, but I also see a lot of other unrelated changes that would break existing functionality. We'd also need to add E2E tests (https://github.com/dyad-sh/dyad/tree/main/e2e-tests) to make sure it works. It might be a bit tricky to test this since it calls an external API (Vercel), but we could create a fake Vercel server, similar to what we do for the AI calls in https://github.com/dyad-sh/dyad/tree/main/testing/fake-llm-server

Let me know if you'd like to make these updates, otherwise i'd be happy to work on adding Vercel support sometime in the future (likely not in the short-term because of other existing feature work), but I think this is a good addition to the overalll Dyad experience.

Thanks again for your time and effort here.

wwwillchen:
@Nokitomo just wanted to let you know that Dyad now natively supports Vercel so I'm going to go ahead and close this PR. Thank you for opening this PR, it was helpful both for inspiration and pointing me towards using Vercel access tokens and I've given you a shout out on the release docs: https://www.dyad.sh/docs/releases/0.13.0-beta.2

PR #782, opened 2025-08-03
fix: improve close button alignment in chat error containers

abdrhxyii:
Adjusted positioning to top-2.5 left-2 and add left padding to prevent
text overlap with the close button. 

image

image


wwwillchen:
thanks!

PR #794, opened 2025-08-04
add editable custom model

B1ueRaw:
Now users can free to edit their custom models by double clicking any custom models created in each provider.
Before this, they have to delete -> create a new one.
I simply add an edit panel (which looks the same as 'Add Custom Model') and integrate that process into the "update" button.

There is one more issue that if a user deletes a model that he was using in chat, then back to chat, that model would still appear (and work) unless user chooses a new one.
Tried to modify "delete-custom-model" in language_model_handlers.ts by the logic that if the name of that model matches the latest using one -> switch to auto (or default) model. Yet I failed, maybe need more explanation for this :)

### DIFF
--- a/e2e-tests/edit_custom_models.spec.ts
+++ b/e2e-tests/edit_custom_models.spec.ts
@@ -1,0 +1,55 @@
+import { test } from "./helpers/test_helper";
+import { expect } from "@playwright/test";
+
+test("edit custom model", async ({ po }) => {
+  await po.setUp();
+  await po.goToSettingsTab();
+  await po.page.getByText("test-provider").click();
+
+  // test edit model by double clicking the model panel
+  await po.page
+    .locator(".text-lg.font-semibold", { hasText: "test-model" })
+    .dblclick({ delay: 100 });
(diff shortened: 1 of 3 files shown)


wwwillchen:
Thanks for opening this PR. Can you add an E2E test to cover this? See https://github.com/dyad-sh/dyad/blob/main/CONTRIBUTING.md#e2e-tests

We already have some tests for this feature area, e.g. https://github.com/dyad-sh/dyad/blob/main/e2e-tests/delete_provider.spec.ts

B1ueRaw:
> Thanks for opening this PR. Can you add an E2E test to cover this? See https://github.com/dyad-sh/dyad/blob/main/CONTRIBUTING.md#e2e-tests
> 
> We already have some tests for this feature area, e.g. https://github.com/dyad-sh/dyad/blob/main/e2e-tests/delete_provider.spec.ts

Two cases were added since I expect users to jump into the edit panel by two ways:
1. Double click the whole custom block area
2. click the edit button

wwwillchen:
LGTM - thanks!

PR #835, opened 2025-08-06
Add support for LLMGateway integration and models

### LINKED ISSUE #845: Add LLMGateway as a provider
Upvote this if you want to see https://llmgateway.io/, an open-source and fully self-hostable alternative to openrouter.

steebchen:
We would love to add support for https://github.com/theopenco/llmgateway, a self-hostable & fully open source LLM Gateway which allows anyone to use any model easily, plus deep insights and observability.

fix #845

### DIFF
--- a/src/components/SetupBanner.tsx
+++ b/src/components/SetupBanner.tsx
@@ -267,7 +267,7 @@ export function SetupBanner() {
                         Setup other AI providers
                       </h4>
                       <p className="text-xs text-gray-600 dark:text-gray-400">
-                        OpenAI, Anthropic, OpenRouter and more
+                        OpenAI, Anthropic, OpenRouter, LLMGateway and more
                       </p>
                     </div>
                   </div>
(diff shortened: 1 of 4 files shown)


wwwillchen:
thanks for opening this PR. I think users can just add llmgateway as a custom provider: https://www.dyad.sh/docs/guides/ai-models/custom-models - i'd prefer to leave it out of the preset list until there's significant community demand (e.g. thumbs up on an existing issue) as it may be confusing to users to see models duplicated across providers

steebchen:
@wwwillchen I just thought it would be nice to have an open-source alternative for openrouter which anyone can just self host by default. but I've created an issue here in case to keep track of interested users: https://github.com/dyad-sh/dyad/issues/845

matbgn:
@wwwillchen you can just prefix model with provider ID as other systems do, what do you think?

It will help anyway to referee provider in case of multiple source for free models for instance

PR #847, opened 2025-08-06
Add missing app description

FoxDavoid:
The description for the program/app is set to the default placeholder. Making it look like [this](https://imgur.com/a/gsIgSx4) in systems like Linux.

So all I did was adding part of the GitHub description as it. Everything else is okay.

Linked image, tho.
Screenshot_20250806_102625

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "dyad",
   "productName": "dyad",
   "version": "0.16.0",
-  "description": "My Electron application description",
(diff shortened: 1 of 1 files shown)


Shaheer-op9872uw:
Makes sense — replacing the placeholder with the GitHub description improves clarity and avoids that generic look on Linux. Everything else checks out.

wwwillchen:
thanks! I didn't realize the description was used 😅

Issue #503, opened 2025-06-27
Claude Code subscription support


swbam:
It would be great if you could login with Claude Code subscription instead of api key.

xied-ctrl:
Yes we want this!

wwwillchen:
If anybody would like to send a PR for this, that would be appreciated! I'm not exactly sure how to integrate with Claude Code since it's a CLI and not an API provider but I've heard of other tools doing this.

IgorWarzocha:
Hi Will, any updates on this? This was the first thing I checked after installing Dyad in my quest to pay less for Claude subscription and use Local LLM for simpler coding, with fallback options for more complex stuff.

Dyad looks REALLY good and I think you're really onto something that Anthropic themselves will replicate eventually.

I know it might not be ultra-helpful (unsure how licensing works), but both Opencode and Gemini CLI (just for Google) have this functionality built in already. This would be a gamechanger.

Keep at it dude. This is looking really promising.

Megajin:
@IgorWarzocha there is already a fork able to do this: https://github.com/ChrisRoyse/CCdyad

Megajin:
@wwwillchen I am pretty sure with IPC this is easy to implement: 

```
    this.claudeProcess = spawn('claude', args, {
      stdio: ['pipe', 'pipe', 'pipe'],
      cwd: this.options.workingDirectory || process.cwd(),
      env: {
        ...process.env,
        ...(this.options.apiKey && { ANTHROPIC_API_KEY: this.options.apiKey })
      }
    });
```

wwwillchen:
@Megajin feel free to open a PR.

Megajin:
@wwwillchen I was cooking: https://github.com/dyad-sh/dyad/pull/865

phatmandrake:
I couldn't tell from #865 if this was decided against or not, it sounded like an architectural decision was being tossed around. I'd be very interested in knowing if anything has been decided one way or another.

wwwillchen:
@phatmandrake - we're definitely open to this, someone identified a clean way of integrating via claude code SDK. if somebody would like to contribute a PR that would be great!

phatmandrake:
> @phatmandrake - we're definitely open to this, someone identified a clean way of integrating via claude code SDK. if somebody would like to contribute a PR that would be great!


https://github.com/zed-industries/agent-client-protocol


Competition is hot right now

JoziGila:
[...]

PR #865, opened 2025-08-07
503 claude code implementation

Megajin:
This will add the new feature #503 . Keep in mind that I am a bit rusty with electron since I have switched to tauri over a year ago. I hope this change will fit in the ecosystem. You don't need to add an API KEY as long as Claude Code is installed and logged in everything will work.

  ## Summary
  • Add comprehensive Claude Code CLI integration as a new model provider
  • Implement secure subprocess management with configurable timeouts
  • Create interactive UI components for settings and permissions
  • Enable real-time streaming and cross-platform compatibility

  ## Key Features
  - **New Model Provider**: "claude-code" provider with direct CLI integration
  - **Process Management**: Secure subprocess handling with configurable timeouts (30-3600 seconds)
  - **Interactive UI**: Complete settings panel, terminal interface, and permission dialogs
  - **Real-time Communication**: Live output streaming and comprehensive error handling
  - **Security**: Permission-based file operations requiring user confirmation
  - **Cross-platform**: Windows/Mac/Linux compatibility with proper path handling

  ## Implementation Details
  - **Provider Integration**: Extended language model system with Claude CLI adapter
  - **UI Components**: Complete React component suite for Claude Code interaction
  - **State Management**: Custom hooks for session and permission management
  - **IPC Architecture**: Full client-server communication layer
  - **Context Transformation**: Converts Dyad context to Claude CLI compatible format
  - **Timeout Configuration**: User-configurable command timeouts with validation

  ## Technical Components
  - **ClaudeCodeModelAdapter**: Bridges Dyad's AI interface with Claude CLI
  - **Security Manager**: Permission system with user approval workflows
  - **Session Manager**: State persistence and process lifecycle management
  - **Process Isolation**: Secure subprocess handling with automatic cleanup
  - **Error Recovery**: Comprehensive error handling and graceful degradation

  ## Test plan
  - [ ] Verify Claude Code provider appears in model selection
  - [ ] Test timeout configuration in settings (30-3600 second range)
  - [ ] Confirm permission dialogs appear for file operations
  - [ ] Validate cross-platform subprocess management
  - [ ] Test error handling and recovery scenarios
  - [ ] Verify context transformation between Dyad and Claude CLI
  - [ ] Test session persistence and cleanup

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dyad",
-  "version": "0.15.0-beta.2",
+  "version": "0.16.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "dyad",
-      "version": "0.15.0-beta.2",
+      "version": "0.16.0",
       "license": "MIT",
       "dependencies": {
         "@ai-sdk/anthropic": "^1.2.8",
--- a/src/__tests__/chat_stream_handlers.test.ts
+++ b/src/__tests__/chat_stream_handlers.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
+import path from "node:path";
 
 import {
   getDyadWriteTags,
@@ -16,6 +17,19 @@ import fs from "node:fs";
 import git from "isomorphic-git";
 import { db } from "../db";
 import { cleanFullResponse } from "@/ipc/utils/cleanFullResponse";
+
+// Mock electron-log to suppress logging during tests
+vi.mock("electron-log", () => ({
+  default: {
+    scope: vi.fn().mockReturnValue({
+      error: vi.fn(),
+      info: vi.fn(),
+      warn: vi.fn(),
+      debug: vi.fn(),
+      log: vi.fn(),
+    }),
+  },
+}));
 
 // Mock fs with default export
 vi.mock("node:fs", async () => {
@@ -56,10 +70,15 @@ vi.mock("isomorphic-git", () => ({
 // Mock paths module to control getDyadAppPath
 vi.mock("../paths/paths", () => ({
   getDyadAppPath: vi.fn().mockImplementation((appPath) => {
-    return `/mock/user/data/path/${appPath}`;
+    return path.join("/mock/user/data/path", appPath);
   }),
   getUserDataPath: vi.fn().mockReturnValue("/mock/user/data/path"),
 }));
+
+// Helper function to normalize paths for cross-platform testing
+const normalizePath = (filePath: string): string => {
+  return path.normalize(filePath).replace(/\\/g, '/');
+};
 
 // Mock db
 vi.mock("../db", () => ({
@@ -696,11 +715,11 @@ describe("processFullResponse", () => {
     });
 
     expect(fs.mkdirSync).toHaveBeenCalledWith(
-      "/mock/user/data/path/mock-app-path/src",
+      path.join("/mock/user/data/path/mock-app-path", "src"),
       { recursive: true },
     );
     expect(fs.writeFileSync).toHaveBeenCalledWith(
-      "/mock/user/data/path/mock-app-path/src/file1.js",
+      path.join("/mock/user/data/path/mock-app-path", "src/file1.js"),
       "console.log('Hello');",
     );
     expect(git.add).toHaveBeenCalledWith(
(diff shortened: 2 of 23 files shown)


wwwillchen:
Thanks @Megajin for opening this PR. I'm not super familiar with Claude Code so please bear with me :)

I patched this PR and tried running it (I logged into Claude Code) but when I try to run Claude Code, I'm getting an error:

```
22:04:33.879 (claude-code-wrapper) › Claude Code failed with exit code 1
22:04:33.879 (claude-code-wrapper) › STDERR: 
22:04:33.879 (claude-code-model-adapter) › Claude Code generation failed: Error: Claude Code failed (exit code 1): 
    at ChildProcess.<anonymous> (/Users/will/Documents/GitHub/dyad-zero/.vite/build/main.js:166709:18)
    at ChildProcess.emit (node:events:518:28)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)
```

Do you know why it's not working?

I took a quick look at the code and it looks like it's basically wrapping around the Claude Code CLI and I'm wondering why you're using this approach vs. using the [Claude Code SDK](https://docs.anthropic.com/en/docs/claude-code/sdk#type-script) directly (i.e. the native TypeScript API). I'm a little concerned that wrapping creates a weird user experience - for example, Dyad revolves around allowing users to approve changes, but with this current approach it looks like Claude is always writing the changes directly. While we do allow users to enable auto approve, by default we require users to approve code changes because Dyad is always running the dev server. This combination of dev server auto-executing code and Claude Code auto-writing code can be quite risky.

I've heard of a lot of good things about Claude Code and I know some users work on their dyad apps alongside with claude code. What's not super clear to me is how much benefit is there for users to have native Claude Code integration inside Dyad? It seems like Claude Code has a powerful Terminal UI (e.g. autocomplete) and Dyad's wrapper around it, will naturally be limited and somewhat jagged.

Let me know what you think - definitely interested in learning more about Claude Code. Thanks.

florianliebig:
One main benefit might be the usage of the claude pro / max plans which include pretty decent amount of usage at a fixed price. You can maybe compare how task-master performs auth (either plain API or Oauth for your subscription): https://github.com/eyaltoledano/claude-task-master/pull/1114 and uses the sdk

fahimanwer:
This is the whole point of adding claude code, to be able to use with Pro/Max Plan, without it it will be better to directly use the API.

Megajin:
Hey guys, I can explain. You always need to install claude-code. Even when using the SDK. You'll need an active subscribtion to make this work and globally install claude-code. I did use this approach because you'll end up with one subscribtion without paying for API-Keys (Tokens basically). The SDK heavily relies on API-Keys and also an installation of claude-code. 

In short: My approach allows the usage of claude-code pro max without paying for more tokens.

Your error most likely comes from a missing claude-code installtion globally or a missing subscribtion. Which again you'll have to login via claud-code cli.

Just another note: claude-flow (https://github.com/ruvnet/claude-flow) uses claude-code in a similiar way. Maybe you'll get some inspiration from there

wwwillchen:
thanks for the feedback everyone - i definitely understand the appeal of using Claude's subscription which is very generous. I think the main question I had was: why not use Claude Code directly instead of through the Dyad UI? Other users have told me they switch back and forth between Dyad and Claude Code on the same project.

@Megajin  I had Claude Code installed but I was using my own API key, I'll try again later after I subscribe.

@florianliebig thanks for the link, that's very helpful - I think if we can use a similar approach, it'll be a much cleaner integration than wrapping the Claude CLI like this

Diegoriggle77:
I'm not home at my laptop I'll check and see what's going sry I'll get to
it asap

ben-vargas:
Saw this because of the mention to task master, I implemented the claude code custom provider there (allowing subscription use), and created the [ai-sdk-provider-claude-code](https://github.com/ben-vargas/ai-sdk-provider-claude-code) - IMO you should absolutely be using the SDK rather than trying to wrap the CLI or spawning Claude Code subprocesses. Hope this helps.

ben-vargas:
> The SDK heavily relies on API-Keys and also an installation of claude-code.

@Megajin - This is not true, SDK works just fine with Max subscriptions and doesn't require API Keys - API Key is just one method of using CC SDK, but the subscription OAuth workflow works equally as well.

Megajin:
> > The SDK heavily relies on API-Keys and also an installation of claude-code.
> 
> @Megajin - This is not true, SDK works just fine with Max subscriptions and doesn't require API Keys - API Key is just one method of using CC SDK, but the subscription OAuth workflow works equally as well.

Ok, looks like I have to do some reading and research. Thanks for pointing that out.

wwwillchen:
thanks for chiming in @ben-vargas - your claude code provider definitely looks promising. we're trying to upgrade to AI SDK v5 #929 but blocked on a dependency.

@Megajin i'm going to close this in favor of the provider approach since it would be cleaner but thanks again for working on this and kickstarting a great discussion.

ben-vargas:
@wwwillchen - Yeah, you don't necessarily need to create a dependency on `ai-sdk-provider-claude-code` unless there are benefits and you're using AI SDK. I also don't know how well it will work with tool calling that might be required looking at dyad. It would need some thorough testing/validation.

Just wanted to share that Claude Code SDK is definitely the way to go probably, over spawning processes... and it would achieve the goal of allowing subscription use of sonnet/opus without an API key which was our main driver in task master.

Megajin:
> thanks for chiming in @ben-vargas - your claude code provider definitely looks promising. we're trying to upgrade to AI SDK v5 #929 but blocked on a dependency.
> 
> @Megajin i'm going to close this in favor of the provider approach since it would be cleaner but thanks again for working on this and kickstarting a great discussion.

Sure go ahead! @ben-vargas thanks again for pointing the sdk out. I have learned something new and I was wrong, thanks again.

PR #911, opened 2025-08-12
[UI] Hide non-setup models in the model picker

Codycody31:
Just hides any non-setup models, given you can't use them properly in the first place.

### DIFF
--- a/src/components/ModelPicker.tsx
+++ b/src/components/ModelPicker.tsx
@@ -16,7 +16,7 @@ import {
   DropdownMenuSubTrigger,
   DropdownMenuSubContent,
 } from "@/components/ui/dropdown-menu";
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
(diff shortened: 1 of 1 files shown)


wwwillchen:
Thanks for opening a PR and taking the time to contribute. I definitely understand the intent of simplifying the UI, but IMO this is fairly complicated and I don't want to change this because there's some complicated and subtle logic here:

- Auto is a special case, where the auto provider is actually Dyad Pro. Whenever Dyad Pro is enabled, all the preset (i.e. cloud type models) are setup.
- The auto model (which can be used when Dyad Pro is disabled) fallbacks to the big 3 AI providers and it should always be shown.
- Some custom models don't actually need an API key.
- There's some interesting UX questions about what if a user selects an OpenAI model but then deletes their API key, IMO, it's weird to hide this from the UI even if it's technically not available.

I'd recommend opening an issue in the future to discuss the change before a PR to save time.

PR #941, opened 2025-08-14
Shard E2E tests

wwwillchen:


Zaid-maker:
Why not migrate to Bun? it's very FAST

PR #958, opened 2025-08-16
Support next.js for routes and handle long width address bar

wwwillchen:


Shaheer-op9872uw (commented):
Well these are apporipiate

Issue #944, opened 2025-08-14
Date/Time Stamps On Messages In The Prompt


beaugast:
Can we get date and time stamps on each message we send in the prompt window? There are times after several prompts that I'm like "this ain't working" and I go to the versions to try to figure out which one I need to restore to before I created the mess I created and I can't seem to reconcile the two.

beaugast:
Or maybe even a number that corresponds to the version if that would be easier?

princeaden1:
Hi @beaugast I can work on this if you can assign me to add the timestamp to the chat.

wwwillchen:
@princeaden1 we can actually show the version associated with each message. We store the commit hash already for each message in the DB https://github.com/dyad-sh/dyad/blob/b06f658fc5ff66f03335a2cb47bfdfaf566f6cfc/src/db/schema.ts#L51

We should be able to get the commit name & timestamp from `useVersions` e.g. https://github.com/dyad-sh/dyad/blob/b06f658fc5ff66f03335a2cb47bfdfaf566f6cfc/src/components/chat/VersionPane.tsx#L29

i'll go ahead and assign this to you

wwwillchen:
thanks @princeaden1 for implementing this!

PR #959, opened 2025-08-16
feat: add timestamp and message version to prompt (#944)

princeaden1:
### Summary

This PR implements a timestamp feature for messages in the prompt window, responding to feature request #944.

**What this does:**
- Prefixes each sent message with a timestamp and message version.
### Screenshot

image

### DIFF
--- a/src/components/chat/ChatMessage.tsx
+++ b/src/components/chat/ChatMessage.tsx
@@ -5,7 +5,12 @@ import {
 } from "./DyadMarkdownParser";
 import { motion } from "framer-motion";
 import { useStreamChat } from "@/hooks/useStreamChat";
-import { CheckCircle, XCircle } from "lucide-react";
+import { CheckCircle, XCircle, Clock, GitCommit } from "lucide-react";
+import { formatDistanceToNow, format } from "date-fns";
+import { useVersions } from "@/hooks/useVersions";
+import { useAtomValue } from "jotai";
+import { selectedAppIdAtom } from "@/atoms/appAtoms";
+import { useMemo } from "react";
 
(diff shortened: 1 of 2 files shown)


src/components/chat/ChatMessage.tsx
```diff
@@ -14,13 +19,44 @@ interface ChatMessageProps {
...
   const { isStreaming } = useStreamChat();
+  const appId = useAtomValue(selectedAppIdAtom);
+  const { versions: liveVersions } = useVersions(appId);
+  // Find the version that was active when this message was sent
+  const messageVersion = useMemo(() => {
+    if (!liveVersions.length || !message.createdAt) return null;
+    const messageTimestamp = new Date(message.createdAt).getTime() / 1000; // Convert to Unix timestamp to compare with version timestamps
+    if (message.role === "user") {
+      const sortedVersions = [...liveVersions].sort(
+        (a, b) => b.timestamp - a.timestamp,
```
wwwillchen: thanks for opening this PR!

we shouldn't need to do timestamp comparison, the assistant message itself has a commitHash.

Also, it's a bit unintuitive in terms of the UX  to show the commit hash as part of the user message, the version + timestamp should be shown as part of the assistant message.

finally, i'd prefer if it's always shown and not only when hovered to make this feature more accessible.

princeaden1:
@wwwillchen Thanks for the thoughtful feedback 🙏 I've pushed the update addressing your points, version + timestamp are now always shown on assistant messages.

image

wwwillchen:
@princeaden1 - thanks, looks great!

Issue #710, opened 2025-07-25
Support Azure AI models


tommasomattarelli:
## Bug Description  
Trying to use a custom model in Dyad with Azure OpenAI (via a custom provider), but I get a `404 Resource Not Found` error when running a prompt. The model setup seems correct on the Azure side.

## Steps to Reproduce  
1. Go to **Settings → Models → Add custom provider**  
2. Configure with:
   - Base URL: `https://<your-resource-name>.openai.azure.com/`
   - API Key: `<your-api-key>`
   - Model: `gpt-4o-mini`
   - Deployment name: (correct deployment from Azure)
3. Save and try to use the model in chat

## Expected Behavior  
The custom model should respond correctly, as configured via Azure OpenAI.

## Actual Behavior  
Dyad throws an error on usage. The model doesn’t respond, and the following 404 error appears in logs.

## System Information
- Dyad Version: 0.13.0
- Platform: win32
- Architecture: x64
- Node Version: v22.14.0
- PNPM Version: 10.13.1
- Node Path: C:\Program Files\nodejs\node.exe
- Telemetry ID: de91f9cf-0603-4097-ae0a-c0c85782049e
- Model: custom::azure:gpt-4o-mini | customId: 10

## Logs
```
  VITE v6.3.4  ready in 449 ms


  ➜  Network: http://192.168.190.50:32100/
  ➜  Network: http://172.23.192.1:32100/

[2025-07-25 09:22:40.373] [error] (chat_stream_handlers) Error streaming text: {
  error: 'AI_APICallError: Resource not found\n' +
    '    at C:\\ProgramData\\tommasomattarelli\\dyad\\app-0.13.0\\resources\\app.asar\\.vite\\build\\main.js:123:4881\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n' +
    '    at async fgt (C:\\ProgramData\\tommasomattarelli\\dyad\\app-0.13.0\\resources\\app.asar\\.vite\\build\\main.js:123:3665)\n' +
    '    at async f2t.doStream (C:\\ProgramData\\tommasomattarelli\\dyad\\app-0.13.0\\resources\\app.asar\\.vite\\build\\main.js:890:15195)\n' +
    '    at async fn (C:\\ProgramData\\tommasomattarelli\\dyad\\app-0.13.0\\resources\\app.asar\\.vite\\build\\main.js:129:19080)\n' +
[...]

PR #1001, opened 2025-08-19
Add Azure OpenAI Custom Model Integration

tmaasen:
Addresses #710 

  This PR implements comprehensive Azure OpenAI integration for Dyad, enabling users to leverage Azure
  OpenAI models through proper environment variable configuration. The implementation adds Azure as a
  supported provider with full integration into the existing language model architecture, including support
  for GPT-5 models. Key features include environment-based
  configuration using `AZURE_API_KEY` and `AZURE_RESOURCE_NAME`, specialized UI components that provide clear
  setup instructions and status indicators, and seamless integration with Dyad's existing provider system.
  The Azure provider leverages the @ai-sdk/azure package (v1.3.25) for compatibility with the current
  TypeScript language model interfaces.

  The implementation includes robust error handling for missing configuration, comprehensive test coverage
  with 9 new unit tests covering critical functionality like model client creation and error scenarios, and
  an E2E test for the Azure-specific settings UI. 

Screenshot 2025-08-18 at 9 14 32 PM


src/ipc/shared/language_model_helpers.ts
```diff
@@ -211,13 +211,56 @@ export const MODEL_OPTIONS: Record<string, ModelOption[]> = {
...
+      temperature: 0,
+    },
+    {
+      name: "gpt-35-turbo",
+      displayName: "GPT-3.5 Turbo",
+      description: "Azure OpenAI GPT-3.5 Turbo model",
+      maxOutputTokens: 4_000,
+      contextWindow: 16_000,
+      temperature: 0,
+    },
```
wwwillchen: I would remove these older models and replace them with the newer GPT 5 models

src/ipc/shared/language_model_helpers.ts
```diff
@@ -258,6 +301,12 @@ export const CLOUD_PROVIDERS: Record<
     websiteUrl: "https://academy.dyad.sh/settings",
     gatewayPrefix: "dyad/",
   },
+  azure: {
+    displayName: "Azure OpenAI",
+    hasFreeTier: false,
+    websiteUrl: "https://portal.azure.com/",
+    gatewayPrefix: "azure/",
```
wwwillchen: this should be a blank string (same as openai), it's a bit complicated, but internally (which is what this is used for), we map openai and azure together.

package.json
```diff
@@ -84,7 +84,8 @@
     "vitest": "^3.1.1"
   },
   "dependencies": {
-    "@ai-sdk/anthropic": "^2.0.4",
+    "@ai-sdk/anthropic": "^1.2.8",
```
wwwillchen: leave anthropic alone

package.json
```diff
@@ -84,7 +84,8 @@
...
+    "@ai-sdk/azure": "^1.3.25",
```
wwwillchen: i recently upgraded to AI sdk v5 so you'll need to bump this version (and retest things work still)

src/__tests__/language_model_helpers_azure.test.ts
wwwillchen, line 1: let's remove this unit test file - this does a lot of mocking and will make future refactorings harder.
most this should already be covered by the e2e test

src/__tests__/get_model_client_azure.test.ts
wwwillchen, line 1: let's remove this unit test file - this does a lot of mocking and will make future refactorings harder.
most this should already be covered by the e2e test

e2e-tests/azure_provider_settings.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  await expect(
+    po.page.getByText("Set these environment variables before starting Dyad"),
+  ).toBeVisible();
+
+  // Check that status indicators show "Not Set" (since no env vars are configured in test)
+  const statusElements = po.page.locator(".bg-red-100, .bg-red-900");
+  await expect(statusElements.first()).toBeVisible();
+});
+
+test("Azure provider appears in provider list", async ({ page }) => {
```
wwwillchen: remove this, it's already covered in test above

src/lib/schemas.ts
```diff
@@ -59,6 +60,7 @@ export type LargeLanguageModel = z.infer<typeof LargeLanguageModelSchema>;
  */
 export const ProviderSettingSchema = z.object({
   apiKey: SecretSchema.optional(),
+  endpoint: SecretSchema.optional(), // For Azure resource endpoint
```
wwwillchen: it doesn't look like this is used - can we remove it?

src/components/settings/ApiKeyConfiguration.tsx
```diff
@@ -46,6 +46,99 @@ export function ApiKeyConfiguration({
...
   isDyad,
 }: ApiKeyConfigurationProps) {
+  // Special handling for Azure OpenAI which requires environment variables
+  if (provider === "azure") {
+    const azureApiKey = envVars["AZURE_API_KEY"];
+    const azureResourceName = envVars["AZURE_RESOURCE_NAME"];
+
+    const isAzureConfigured = !!(azureApiKey && azureResourceName);
+
+    return (
```
wwwillchen: this should be extracted into a separate component and ideally separate file

src/components/settings/ApiKeyConfiguration.tsx
```diff
@@ -46,6 +46,99 @@ export function ApiKeyConfiguration({
...
+      <div className="space-y-4">
+        <Alert variant={isAzureConfigured ? "default" : "destructive"}>
+          <Info className="h-4 w-4" />
+          <AlertTitle>Azure OpenAI Configuration</AlertTitle>
```
wwwillchen: this UI is almost unreadable in dark mode btw

Image

wwwillchen (changes requested):
thanks for opening the PR!

our project relies primarily on e2e testing than unit tests (although we do have a few) to make it easier to refactor. since azure has some special handling around env vars, i recommend removing the unit tests and instead creating an e2e test that actually sends a prompt / chat message through to azure and updating the e2e test server to support azure: https://github.com/dyad-sh/dyad/blob/34215db141ddb42495325aebc644d17a509c9754/testing/fake-llm-server/index.ts#L140 

you can configure a custom azure baseUrl in get_model_client to point to the e2e test server (this is what we use for some of the other providers)

src/components/settings/ApiKeyConfiguration.tsx
tmaasen: Good catch, resolved

image

tmaasen:
@wwwillchen should be good now. First time doing e2e tests with playwright, so it took me a minute to see how it all worked. Very neat setup with the fake llm server and all!

e2e-tests/snapshots/azure_send_message.spec.ts_send-message-through-Azure-OpenAI-1.aria.yml
```diff
@@ -0,0 +1,6 @@
+- paragraph: hi
+- paragraph: undefinedundefinedundefinedundefinedundefinedundefinedundefinedundefinedundefinedundefined
```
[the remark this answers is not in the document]
wwwillchen: @tmaasen I think you need to fix the e2e test mocks for azure - somthing seems off

package-lock.json
```diff
@@ -175,13 +211,13 @@
       }
     },
```
wwwillchen: not sure why this is changing, but it might be causing the TS compiler failure in CI

wwwillchen:
@tmaasen thanks for the changes, i think once the e2e test is fixed + GitHub action check failure is fixed (see: https://github.com/dyad-sh/dyad/actions/runs/17114793652/job/48721364589?pr=1001), we can merge this.

package-lock.json
tmaasen: Looks like it changed to 2.0.17 because @ai-sdk/azure v2.0.17 has a direct dependency for @ai-sdk/openai to be v2.0.17

Also in the diff, I see my PR is removing the "-beta.1" from the version. I checked against upstream/main, and the package.json has "0.18.0" specified for the version, but the package-lock.json has the "-beta.1", https://github.com/dyad-sh/dyad/blob/main/package-lock.json. Do you want me to keep the "-beta.1" as-is?

tmaasen:
@wwwillchen FWIW, I did want to make you aware that Microsoft's recommended approach for integrating custom azure openai models is using the `@azure/` packages rather than using `@ai-sdk/` packages (see [Azure AI Foundry package.json example](https://github.com/azure-ai-foundry/foundry-samples/blob/main/samples/microsoft/typescript/mslearn-resources/quickstart/package.json) and [MS Docs](https://learn.microsoft.com/en-us/azure/ai-foundry/quickstarts/get-started-code?tabs=typescript&pivots=fdp-project) and [Azure AI Foundry Docs](https://ai.azure.com/doc/azure/ai-foundry/how-to/develop/sdk-overview)). 

I tried using the `@ai-sdk/azure` because it aligned more with dyad's AI usage approach, and while they both seem to work, using the `@azure/` packages has a few more auth features for working with private models in azure. Using their packages would've probably meant quite a bit more refactoring of dyad's codebase too.

Just wanted to give you all the info before we merge!

package-lock.json
wwwillchen: hm... it looks like weirdly the bump from 2.0.15 to 2.0.17 did cause the type failure which is very weird since it's a patch version bump (I think AI SDK accidentally regressed their types). I've fixed the version so we can still use 2.0.15 for openai and then let azure use  openai 2.0.17

wwwillchen:
> @wwwillchen FWIW, I did want to make you aware that Microsoft's recommended approach for integrating custom azure openai models is using the `@azure/` packages rather than using `@ai-sdk/` packages (see [Azure AI Foundry package.json example](https://github.com/azure-ai-foundry/foundry-samples/blob/main/samples/microsoft/typescript/mslearn-resources/quickstart/package.json) and [MS Docs](https://learn.microsoft.com/en-us/azure/ai-foundry/quickstarts/get-started-code?tabs=typescript&pivots=fdp-project) and [Azure AI Foundry Docs](https://ai.azure.com/doc/azure/ai-foundry/how-to/develop/sdk-overview)).
> 
> I tried using the `@ai-sdk/azure` because it aligned more with dyad's AI usage approach, and while they both seem to work, using the `@azure/` packages has a few more auth features for working with private models in azure. Using their packages would've probably meant quite a bit more refactoring of dyad's codebase too.
> 
> Just wanted to give you all the info before we merge!

got it, thanks for the details! I think using @ai-sdk is OK because it seems unlikely ppl will need to use private models with Dyad but we can always revisit this in the future.

tmaasen:
@wwwillchen anymore changes requested? I see there are 4 CI (e2e tests) failing, but I see they are failing in a few other PRs too. I tried looking into one of them, but the failures don't seem directly related to this PR's changes.

wwwillchen:
@tmaasen looks like some of the tests are failing because of the change you made to dump path (which I agree makes sense). Could you rebaseline the e2e tests https://github.com/dyad-sh/dyad/blob/main/CONTRIBUTING.md#e2e-tests? thanks

tmaasen:
> @tmaasen looks like some of the tests are failing because of the change you made to dump path (which I agree makes sense). Could you rebaseline the e2e tests https://github.com/dyad-sh/dyad/blob/main/CONTRIBUTING.md#e2e-tests? thanks

@wwwillchen been looking at this for a bit, and for now I am just reverting my changes to `chatCompletionHandler`. There were some edge cases with how some of the e2e tests handled/expected dumps, and it's not directly related to this feature. Once I reverted, the small pool of e2e tests I was using locally (ones that failed in the CI) are now passing, so we should be in good shape.

I kept the changes to `chatCompletionHandler` as a stash, so I can work on it in another bug in the future if you think it's still a worthy improvement.

wwwillchen (approved):
Thanks for this PR! Merging it in now.

Issue #1072, opened 2025-08-25
Long words overflow outside of the chat container


azizmejri1:
**Describe the bug**
Long words in chat messages do not wrap and instead overflow outside the chat container.

**To Reproduce**
1. Open the chat interface
2. Send a message with a very long word (e.g. "Supercalifragilisticexpialidociousssssssssssssssssssssss")
3. Observe that the word overflows beyond the message container

**Expected behavior**
Long words should wrap to the next line instead of overflowing outside the chat message.

**Screenshots**

<img width="558" height="312" alt="Image" src="https://github.com/user-attachments/assets/c353410f-22fc-44c9-821e-792fc0339fbb" />

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser N/A
- Version N/A

**Additional context**
N/A

PR #1073, opened 2025-08-25
fix text overflow in chat messages

azizmejri1:
This PR fixes the issue #1072 where long words in chat messages overflow outside the message container.  
The fix applies Tailwind’s `break-words` utility to ensure that overly long words wrap properly within the chat bubble.
    

---

## Summary by cubic
Fixes text overflow in chat messages by adding Tailwind’s break-words to the message content. Long words and URLs now wrap inside the bubble instead of spilling out.


wwwillchen:
Thanks!

PR #1076, opened 2025-08-25
Prompts: align examples to TypeScript (.tsx/.ts) to avoid .jsx drift

antont:
The prompt text says:

> - Use TypeScript.

But in the examples there are .jsx files.

I got errors sometimes (in my fork with other changes too) that JSX got generated instead of TSX. This seems to have fixed it for me.
    

---

## Summary by cubic
Updated system prompt examples to use .tsx/.ts instead of .jsx/.js, matching the “Use TypeScript” instruction. This prevents JSX drift and reduces cases where generated code uses JSX instead of TSX.


wwwillchen:
thanks! this makes sense. I need to update our test baselines and then will merge this in.

PR #1116, opened 2025-08-28
search chat feature

iamevansobeng:
created the feature to search for chats in an app by the chat title or the messages in the individual chats
    

---

## Summary by cubic
Adds chat search in the sidebar. Users can find chats by title or message content within the selected app.

- **New Features**
  - Search input in ChatList; sends query to useChats(appId, query).
  - New IPC handler chat:search; searches titles and message content, merges and dedupes results scoped to appId.
  - IpcClient.searchChats and lib/chat.searchChats added; preload allows chat:search.
  - E2E test covers the end-to-end search flow.


e2e-tests/chat_search.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { expect } from "@playwright/test";
+import { test } from "./helpers/test_helper";
+test("user can search for chats in an app", async ({ po }) => {
+  // Go to the chat tab (or wherever your chat sidebar is)
+  await po.goToChatTab();
+
+  // Optionally, create a new chat with a known title for testing
```
wwwillchen: I would create multiple chats to make sure the logic and ordering is as expected, you can rename the chat (i think you'll need to create page object helpers to do this)

e2e-tests/chat_search.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  await po.clickNewChat();
+  // Add a delay or wait for the chat to appear if needed
+
+  // Fill the search input (use getByPlaceholder or getByTestId for reliability)
+  const searchInput = po.page.getByPlaceholder("Search chats...");
+  await searchInput.fill("New Chat");
+
+  // Wait for the search results to update
+  await po.page.waitForTimeout(500); // or use a more robust wait
```
wwwillchen: i'd use a more robust wait (e.g. wait for a specific element to appear)

e2e-tests/chat_search.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  // Assert that the chat list contains the expected result
+  const chatListItems = await po.page
+    .locator(".chat-list-item")
+    .allTextContents();
```
wwwillchen: I'd think about just snapshotting the entire chat list container, e.g. https://github.com/dyad-sh/dyad/blob/6e4eb7c46b1a274410b3fcd10a47063e68e5a062/e2e-tests/helpers/test_helper.ts#L453

e2e-tests/chat_search.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  // Fill the search input (use getByPlaceholder or getByTestId for reliability)
+  const searchInput = po.page.getByPlaceholder("Search chats...");
+  await searchInput.fill("New Chat");
+
+  // Wait for the search results to update
+  await po.page.waitForTimeout(500); // or use a more robust wait
+
+  // Assert that the chat list contains the expected result
+  const chatListItems = await po.page
+    .locator(".chat-list-item")
```
wwwillchen: generally, i prefer to use test id since it's more robust (and in case we refactor the css/class names, it won't break the test): https://playwright.dev/docs/locators#locate-by-test-id

wwwillchen (commented):
Thanks for the PR! This is a good start.

Overall feedback:
- I tested it out locally and it does work but I think the UX would be better having the search open a dialog similar to ChatGPT because there's not a lot of space in the current chat sidebar to display the chat title. Also, I think we should display the part of the chat message that matched as well (if any, otherwise perhaps just show the beginning of the first message)

Image

- for the chats that match, we should sort by newest chat first since those are generally what users want to see first (I think chatgpt also does this).
- i noticed a somewhat jarring / flickering with the loading UI as I was typing, I think we should just remove the loading UI (since it's usually very fast since it's all local) or do some kind of debouncing with the search input.

Also, I added instructions for the presubmit hook (https://github.com/dyad-sh/dyad/blob/main/CONTRIBUTING.md#setup) so you can catch formatting issues locally. I've also given you triage access to the repo so you should be able to run the GitHub Actions workflow (e.g. tests) yourself

Thanks!

iamevansobeng:
Thanks @wwwillchen for the detailed feedback and for testing this locally! Really appreciate the constructive suggestions.

I agree with your points, especially about the UX. You're absolutely right that the current sidebar has limited space for displaying full chat titles. It escaped me while working. I'll work on implementing a dialog-based search similar to ChatGPT that can show:
- Full chat titles
- Matching message snippets (or beginning of first message if no content match)

For the other points:
- **Sorting**: Will implement newest-first ordering ✅
- **Loading UI**: I'll add debouncing to the search input to reduce the flickering effect
- **Tests**: Will refactor to use test IDs, create multiple test chats, and implement proper waits instead of timeouts

I'll also address the bot's feedback regarding:
- Input validation for appId 
- Race condition handling in the useEffect
- Proper error handling

Thanks for the triage access and the presubmit hook info - I'll get that set up to catch formatting issues locally.

Let me work on these improvements and push an updated version. Should I open this as a draft PR while I implement the dialog-based search, or would you prefer I keep it as-is and iterate?

wwwillchen:
ok great, feel free to use this same PR, thank

PR #1139, opened 2025-08-30
Component selection shortcut

azizmejri1:
This PR introduces a new keyboard shortcut to improve the efficiency of selecting components in the app. Users can now quickly select components using Meta + Shift + C for Mac and Ctrl + Shift + C for Other devices (Windows/Linux)

    

---

## Summary by cubic
Add a shortcut to quickly activate the component selector from the preview. Use Meta+Shift+C on macOS and Ctrl+Shift+C on Windows/Linux.

- **New Features**
  - Added useShortcut hook to handle key combos and prevent default on match.
  - Wired shortcut in PreviewIframe with OS detection for Meta vs Ctrl.
  - Forwarded keydown events from the iframe to the parent via postMessage (dyad-shortcut-triggered) so the shortcut works inside preview content.


src/components/preview_panel/PreviewIframe.tsx
wwwillchen, line 1: In L443 - let's also update the tooltip to display the keyboard shortcut (depending on platform) so that users know this exists. thanks!

azizmejri1: Thanks for the suggestion . I fixed the issues raised by the bot and updated the tooltip to display the keyboard shortcut.

worker/dyad-component-selector-client.js
```diff
@@ -146,6 +146,37 @@
...
+    const hasShift = e.shiftKey;
+    const hasCtrlOrMeta = e.ctrlKey || e.metaKey;
+    if (key && hasShift && hasCtrlOrMeta) {
+      e.preventDefault();
+      window.parent.postMessage(
+        {
+          type: "dyad-shortcut-triggered",
+          key: key,
+          eventModifiers: {
+            ctrl: e.ctrlKey,
```
wwwillchen: Just to be safe, rather than posting a generic "dyad-shortcut-triggered" message, I would post a specific message like "dyad-select-component-shortcut" and not send the actual key (or event modifiers) to the parent window. 

See the graphite-app security comment for more context.

src/components/preview_panel/PreviewIframe.tsx
```diff
@@ -433,6 +446,7 @@ export const PreviewIframe = ({ loading }: { loading: boolean }) => {
                     ? "Deactivate component selector"
                     : "Select component"}
                 </p>
+                <p>{isMac ? "Meta + Shift + C" : "Ctrl + Shift + C"}</p>
```
wwwillchen: let's use ⌘ for Meta and ⇧ for Shift (this is more idiomatic and what Chrome DevTools does)

wwwillchen (changes requested):
Just pulled it down and it works very nicely! Couple of suggestions around security and tooltip and then I think we can merge. thank you

wwwillchen (approved):
awesome! thank you.

PR #1140, opened 2025-08-30
Codebase Improvements: Documentation Fix, Function Parsing, and Telemetry Cleanup

rakibulrocky14:
## Summary

- **Contributor Guide**  
  Corrected a typo so that instructions now reference a pull “request” rather than a “requet”.

- **Function Name Parsing**  
  Improved `getFunctionNameFromPath` to correctly return the basename for both direct UI file paths and index files, and exported it for test coverage.

- **TelemetryBanner Cleanup**  
  Removed outdated TODO comments and relied on existing state to hide the banner once consent is set.

- **Unit Test Coverage**  
  Added unit tests for directory, file, and index scenarios to ensure function-name parsing works correctly and to prevent regressions.
    

---

## Summary by cubic
Fixes function-name parsing for files, directories, and index paths, and cleans up telemetry banner logic and docs. Adds unit tests to prevent regressions.

- **Bug Fixes**
  - getFunctionNameFromPath now returns the correct name for files, directories, and index.* paths; exported and covered by unit tests.

- **Refactors**
  - Removed outdated TODOs and use existing state to hide TelemetryBanner after consent.
  - Fixed CONTRIBUTING typo (“requet” → “request”).


rakibulrocky14:
@wwwillchen

rakibulrocky14:
@wwwillchen

src/__tests__/getFunctionNameFromPath.test.ts
```diff
@@ -0,0 +1,16 @@
+import { getFunctionNameFromPath } from "@/ipc/processors/response_processor";
+import { describe, expect, it } from "vitest";
+
+describe("getFunctionNameFromPath", () => {
+  it("returns directory name for index files", () => {
+    expect(getFunctionNameFromPath("/foo/bar/index.ts")).toBe("bar");
+  });
+
```
wwwillchen, line 8: I don't know if this is the right expectation - when i look at supabase edge function examples (e.g. https://github.com/supabase/supabase/tree/master/examples/edge-functions/supabase/functions) it's always function-name/index.ts - is there supabase docs that says edge functions aren't always exported from index.ts?

rakibulrocky14: @wwwillchen 

Supabase’s official examples and CLI scaffolding do indeed use the function-name/index.ts convention, and the code fully supports that layout.

I added a small safeguard so that if someone provides a direct file path (e.g., hello.ts) the function still works, even though this pattern isn’t explicitly documented. It doesn’t change the behavior for the standard folder/index.ts structure, so existing workflows remain unaffected.

If the project wants to enforce the official convention strictly, I can revert to the narrower implementation, but keeping the broader support is harmless and future‑proofs the tool. 

> Which approach do you prefer? Please let me know.

wwwillchen: let's stick with the official convention. we can do broader support later if we get users requesting this. thanks

rakibulrocky14:
@wwwillchen

wwwillchen:
@rakibulrocky14 - as mentioned in earlier comment, i think we should just continue with the index.ts convention for edge functions. i think some of the other stuff has already landed in earlier PRs so I'll go ahead and close this.

Issue #945, opened 2025-08-15
<dyad-command type="refresh"></dyad-command> appearing in generated text output


princeaden1:
**Describe the bug**
While using the application, the literal string <dyad-command type="refresh"></dyad-command> is sometimes included in generated text instead of being interpreted or removed. This looks like an internal control/command tag that is not meant to be shown to end users.

**To Reproduce**

Ask Dyad to generate a code, at the bottom and above the summary it use to include the tag asking user to refresh

**Expected behavior**
While using the application, the literal string <dyad-command type="refresh"></dyad-command> is sometimes included in generated text instead of being interpreted or removed. This looks like an internal control/command tag that is not meant to be shown to end users.

**Screenshots**

<img width="814" height="200" alt="Image" src="https://github.com/user-attachments/assets/721bc90c-4dab-4aaf-8b9c-51b88c2a7e26" />

**Desktop (please complete the following information):**

**Additional context**

wwwillchen:
yeah, we should hide it

PR #1162, opened 2025-09-02
fix: hide <dyad-command> tags from generated text output (#945)

princeaden1:
### Summary
This PR hides `<dyad-command type="refresh"></dyad-command>` tags from generated text outputs. These elements are not intended for end users and clutter the response text.

### Changes
- Filter out `<dyad-command>` tags from generated text.
- Ensure no regressions in rendering or formatting of user-visible output.

    

---

## Summary by cubic
Hide dyad-command tags from chat output so internal commands never appear in user-visible text. Updates DyadMarkdownParser to detect and drop these tags during preprocessing, parsing, and rendering.

### DIFF
--- a/src/components/chat/DyadMarkdownParser.tsx
+++ b/src/components/chat/DyadMarkdownParser.tsx
@@ -123,6 +123,7 @@ function preprocessUnclosedTags(content: string): {
     "dyad-edit",
     "dyad-codebase-context",
     "think",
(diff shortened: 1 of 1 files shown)


rakibulrocky14:
This is really an important UI reated PR because this tag shows in the output test

wwwillchen:
@princeaden1 thanks!

Issue #1194, opened 2025-09-05
No resizing limit


azizmejri1:
**Describe the bug**
Currently, Dyad windows can be resized without any constraints. This allows the window to shrink to unusably small sizes

**To Reproduce**

1. Open Dyad.  
2. Resize the window continuously.  
3. Notice there are no restrictions on minimum size. 

**Expected behavior**
- Set a minimum window width and height to prevent UI elements from becoming inaccessible.  

**Screenshots**

https://github.com/user-attachments/assets/54e843c0-2abb-463a-bd19-16b35a1e3ec2

**Desktop (please complete the following information):**

- OS: Windows 10

**Additional context**
N/A

PR #1195, opened 2025-09-05
Set minimum window size to prevent UI breakage

azizmejri1:
This PR solves issue #1194 by setting a minimum height and width
    

---

## Summary by cubic
Set a minimum window size to prevent UI breakage when the app is resized too small. The BrowserWindow now enforces minWidth 800 and minHeight 500 so layouts stay stable and controls remain accessible.


wwwillchen:
thanks!

PR #1201, opened 2025-09-05
fix(ui): turbo edit -  prevent code editor content click from toggling accordion

princeaden1:
Hovering over the turbo edit code block previously showed a pointer cursor, and clicking inside would collapse/expand the accordion. This PR updates the DyadEdit component to use cursor-text and adds e.stopPropagation() so clicks inside the code block no longer toggle the accordion.

https://github.com/user-attachments/assets/1b23be66-6d8f-4e4b-a390-d59f9a2bcee8

    

---

## Summary by cubic
Prevented the Turbo Edit accordion from toggling when clicking inside the code block and updated the cursor to text to signal editing.

- **Bug Fixes**
  - Stop click propagation inside the code block to avoid accordion toggle.
  - Apply cursor-text on the code container for proper text-selection affordance.

### DIFF
--- a/src/components/chat/DyadEdit.tsx
+++ b/src/components/chat/DyadEdit.tsx
@@ -99,7 +99,10 @@ export const DyadEdit: React.FC<DyadEditProps> = ({
         </div>
       )}
       {isContentVisible && (
-        <div className="text-xs">
+        <div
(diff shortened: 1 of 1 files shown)


wwwillchen:
thanks!

PR #1259, opened 2025-09-11
Make Azure OpenAI Provider Configuration UI-Only

tmaasen:
Previously: configured AZURE_API_KEY and AZURE_RESOURCE_NAME by setting them in the terminal before running.

Now: Set them directly within the UI settings.
    

---

## Summary by cubic
Moves Azure OpenAI configuration to UI-only. Users now enter Azure Resource Name and API Key in Settings; environment variables are no longer used.

- **New Features**
  - Added Azure settings form with Resource Name and API Key fields, save, and success/error states.
  - Persist Azure credentials in settings (AzureProviderSetting: azureApiKey, resourceName).
  - Provider status now checks settings, not env vars.
  - Model client initializes Azure via createAzure(...) using saved settings, with clearer missing-config errors.

- **Migration**
  - Stop using AZURE_API_KEY and AZURE_RESOURCE_NAME environment variables.
  - Go to Settings → Providers → Azure and save Resource Name and API Key.
  - If you previously relied on env vars, Azure will appear unconfigured until saved in the UI.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dyad",
-  "version": "0.20.0",
+  "version": "0.21.0-beta.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "dyad",
-      "version": "0.20.0",
+      "version": "0.21.0-beta.1",
       "license": "MIT",
       "dependencies": {
         "@ai-sdk/amazon-bedrock": "^3.0.15",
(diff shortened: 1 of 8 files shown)


Diegoriggle77:
Output has to be:  Triple A studio quality to set industry standard.  Accurate and security and privacy non negotiable.

wwwillchen:
Thanks @tmaasen for the PR - I think adding UI as an option is good. Generally for our other providers we look-up UI config values first and then fallback to environment variables. This way users can explicitly put it in the UI but also benefit from implicit env var if it's there.

wwwillchen:
I'm actually going to close this in favor of #1289 which basically implements the suggestion above. Thanks again for the PR!

PR #1289, opened 2025-09-15
Enhance Azure configuration handling and UI updates

rakibulrocky14:
# Changes 

- Update Azure configuration components to manage API key and resource name settings.
- Improve visibility of conThis pull request introduces a redesigned Azure provider settings UI and refactors the logic for configuring Azure OpenAI credentials, both in the frontend and supporting hooks. The main changes focus on making the Azure configuration experience clearer and more robust by supporting both environment variable and saved settings, improving status indicators, and updating the logic for determining provider readiness.

**Azure Provider UI and Logic Improvements**

* Added a new `AzureConfiguration` component that provides a dedicated form for entering and saving Azure resource name and API key, with clear status indicators and error handling. The UI now explains precedence between saved settings and environment variables, and guides users through configuration. (`src/components/settings/AzureConfiguration.tsx`)
* Updated the main provider settings page and API key configuration logic to pass Azure-specific settings and update functions to the new component, ensuring seamless integration and correct state management. (`src/components/settings/ApiKeyConfiguration.tsx`, `src/components/settings/ProviderSettingsPage.tsx`) [[1]](diffhunk://#diff-2104fb487cda3768cc5777889100e882f51e7fb3e13abe3cc89cf8ed1444300aR35) [[2]](diffhunk://#diff-2104fb487cda3768cc5777889100e882f51e7fb3e13abe3cc89cf8ed1444300aR51-R61) [[3]](diffhunk://#diff-9140e707ebb56ffed3272b4661ea1e6d8388ee604a8535c58e8a1564d280057cR297)
* Refactored the logic for determining whether Azure is configured to check both saved settings and environment variables, ensuring accurate status display and enabling fallback to environment variables if no settings are saved. (`src/components/settings/ProviderSettingsPage.tsx`, `src/hooks/useLanguageModelProviders.ts`) [[1]](diffhunk://#diff-9140e707ebb56ffed3272b4661ea1e6d8388ee604a8535c58e8a1564d280057cL72-R99) [[2]](diffhunk://#diff-9ac9e279a0cda34a0bc519348d5474b2e355b0828a678495be3af1e8984b5be5R35-R48)
* Updated the Azure provider E2E test to verify the new UI elements, status indicators, and guidance, ensuring the test matches the new configuration flow and messaging. (`e2e-tests/azure_provider_settings.spec.ts`)

**Supporting Type and Import Updates**

[...]

PR #1303, opened 2025-09-17
Fix/azure configuration

dvansari65:
Added: added input fields and save keys button with form submission.
Created onConfigSave method which is taking azure api key and resource name as a input
    

---

## Summary by cubic
Adds an in-app Azure OpenAI configuration form to enter and save the API key and resource name, fixing the missing option when env vars aren’t set. Also bumps Electron to v38 for the desktop app.

- New Features
  - Added input fields and a Save button in Azure settings for AZURE_API_KEY and AZURE_RESOURCE_NAME.
  - Hooked form submission to onConfigSave; shows “Saving…” and disables the button while saving.
  - Status chips now reflect entered values (not only env vars).

- Dependencies
  - Upgraded electron to ^38.1.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dyad",
-  "version": "0.20.0",
+  "version": "0.21.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "dyad",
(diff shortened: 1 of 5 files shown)


wwwillchen:
this is being worked on in #1289

PR #1316, opened 2025-09-18
Fix/tooltip for chat title

dvansari65:
>>Created Tooltip component for scalability.
>>updated chatList component by using tool tip component .
>>Now user can easily see their chat title irrespective size of the title.
>>for lower length string tooltip will not appear since chat title already visible in the chat title

    

---

## Summary by cubic
Adds a reusable tooltip and uses it in ChatList to reveal long chat titles without truncation. Also adds a settings form to enter and save Azure OpenAI API key and resource name.

- New Features
  - Tooltip component; used in ChatList to show full title on hover only when >20 chars.
  - Azure settings: inputs for API key and resource name, Save button (isSaving state), and Set/Not Set indicators.

- Dependencies
  - Electron upgraded from 35.1.4 to ^38.1.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dyad",
-  "version": "0.20.0",
+  "version": "0.21.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "dyad",
-      "version": "0.20.0",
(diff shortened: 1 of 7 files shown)


wwwillchen:
let's scope this PR to only be fixing the chat title.

Issue #1279, opened 2025-09-13
[bug] Ollama models scroll with many models in the list


roman-supy-io:
## Bug Description (required)
Ollama models are not scrollable/visible if there are many models (I have >20)

## Steps to Reproduce (required)
- Install Ollama
- Install at least 15 models (in my case) even if the screen is maximized

## Expected Behavior (required)
The list should be scrollable and the all models are available to choose from

## Actual Behavior (required)
The list is not scrollable and remaining models at the bottom are not reachable

## System Information
- Dyad Version: 0.20.0
- Platform: darwin
- Architecture: arm64
- Node Version: v22.6.0
- PNPM Version: n/a
- Node Path: /Users/roman/.nvm/versions/node/v22.6.0/bin/node
- Telemetry ID: e8d5741d-e967-497d-b9ae-8793e391e033
- Model: <any>

## Logs
```
[2025-09-12 11:38:50.128] [error] (pro_handlers) LLM Gateway API key (Dyad Pro) is not configured.
[2025-09-13 03:19:57.894] [error] (pro_handlers) LLM Gateway API key (Dyad Pro) is not configured.
[2025-09-13 04:01:08.294] [error] (pro_handlers) LLM Gateway API key (Dyad Pro) is not configured.
[2025-09-13 21:27:54.507] [error] (pro_handlers) LLM Gateway API key (Dyad Pro) is not configured.
[2025-09-13 21:29:26.796] [warn]  (runShellCommand) Stderr from "pnpm --version": /Users/roman/.nvm/versions/node/v22.6.0/lib/node_modules/corepack/dist/lib/corepack.cjs:21535
  if (key == null || signature == null) throw new Error(`Cannot find matching keyid: ${JSON.stringify({ signatures, keys })}`);
                                              ^

[...]

PR #1323, opened 2025-09-18
fix(ui): add scrolling to model dropdown when list is long (#1279)

princeaden1:
## Problem
The model selection dropdown doesn't scroll when there are many models available, causing the list to extend beyond the viewport and become unusable.

## Solution
- Added `max-h-100 overflow-y-auto` classes to the DropdownMenuContent

This closes (#1279)

    

---

## Summary by cubic
Adds vertical scrolling to the model selection dropdown so long lists don’t overflow the viewport. Applies max height and overflow-y auto to DropdownMenuSubContent across provider, Ollama, and LM Studio menus to keep the list usable.

### DIFF
--- a/src/components/ModelPicker.tsx
+++ b/src/components/ModelPicker.tsx
@@ -313,7 +313,7 @@ export function ModelPicker() {
                       </span>
                     </div>
                   </DropdownMenuSubTrigger>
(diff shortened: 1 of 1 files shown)


wwwillchen:
thanks!

PR #1338, opened 2025-09-20
[FIX] sanitize GitHub repo names to prevent space-to-dash mismatch

akash1520:
Fixes issue where GitHub automatically converts spaces to dashes in repo names but Dyad continued referencing the original names with spaces, causing connection and push failures.
    

---

## Summary by cubic
Sanitizes GitHub repo and local app folder names (spaces → dashes) to match GitHub behavior. Fixes failures when creating, connecting, or pushing to repos with spaces.

- **Bug Fixes**
  - Added sanitizeAppFolderName and ensureAppFolderSanitized to rename folders, update DB (path/githubRepo), and guard against conflicts.
  - Create repo: always uses sanitized folder name; stores sanitized repo; repo param is now optional.
  - Connect existing: sanitizes repo before lookup and save; clearer error includes attempted sanitized name.
  - Push: sanitizes folder first; normalizes and persists githubRepo before building the remote URL.
  - Updated IPC handler typing for github:create-repo.

### DIFF
--- a/e2e-tests/github.spec.ts
+++ b/e2e-tests/github.spec.ts
@@ -143,3 +143,35 @@ test("create and sync to existing repo - custom branch", async ({ po }) => {
     operation: "create",
   });
 });
+
+test("create repo with spaces in name - should sanitize blank spaces to dashes", async ({
+  po,
+}) => {
+  await po.setUp();
+  await po.sendPrompt("tc=basic");
+
+  await po.getTitleBarAppNameButton().click();
+  await po.githubConnector.connect();
+
+  await expect(po.githubConnector.getCreateNewRepoModeButton()).toHaveClass(
+    /bg-primary/,
+  );
+
+  await po.githubConnector.fillCreateRepoName("test repo with spaces");
+
+  await po.page.waitForSelector("text=Repository name is available!", {
+    timeout: 5000,
+  });
+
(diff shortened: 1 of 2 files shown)


wwwillchen:
thanks for the PR. this fix feels a bit heavy-handed as it's actually **renaming** the folder as part of the git sync (which seems quite surprising from a user perspective). I don't think we need to rename the app folder (IIUC) to fix this issue.

Also, let's add an e2e test https://github.com/dyad-sh/dyad/blob/main/CONTRIBUTING.md#e2e-tests - probably add a test case here: https://github.com/dyad-sh/dyad/blob/main/e2e-tests/github.spec.ts

akash1520:
> thanks for the PR. this fix feels a bit heavy-handed as it's actually **renaming** the folder as part of the git sync (which seems quite surprising from a user perspective). I don't think we need to rename the app folder (IIUC) to fix this issue.
> 
> Also, let's add an e2e test https://github.com/dyad-sh/dyad/blob/main/CONTRIBUTING.md#e2e-tests - probably add a test case here: https://github.com/dyad-sh/dyad/blob/main/e2e-tests/github.spec.ts

@wwwillchen Got rid of folder renaming, plus added an e2e test.

wwwillchen:
thanks for the PR - implemented in https://github.com/dyad-sh/dyad/pull/2377

PR #1341, opened 2025-09-21
implementing app naming feature

azizmejri1:
This PR implements automatic app naming
    

---

## Summary by cubic
Adds automatic app naming from the first chat summary and centralizes rename logic for safer, consistent behavior. Improves UX by naming new apps automatically and reduces duplication.

- **New Features**
  - Auto-rename app to the chat’s <dyad-chat-summary> when it’s the only, untitled chat.
  - Safe rename: locks per app, checks name/path conflicts, stops running app, moves files if needed, rolls back on failure, and updates the DB.

- **Refactors**
  - Extracted rename flow to src/ipc/utils/rename_app.ts and used in handlers.
  - Exported copyDir for reuse.

### DIFF
--- a/src/ipc/handlers/app_handlers.ts
+++ b/src/ipc/handlers/app_handlers.ts
@@ -51,10 +51,11 @@ import { isServerFunction } from "@/supabase_admin/supabase_utils";
 import { getVercelTeamSlug } from "../utils/vercel_utils";
 import { storeDbTimestampAtCurrentVersion } from "../utils/neon_timestamp_utils";
 import { AppSearchResult } from "@/lib/schemas";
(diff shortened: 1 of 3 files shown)


wwwillchen (commented):
thanks for the PR. I think using chat title is clever but it doesn't seem like the right solution because it creates fairly long app names that aren't very intuitive like "Creating Recipe Finder & Meal Planner"

I've been thinking about a different approach where we first generate an "app proposal" (e.g. PRD, app name, color theme) for the first very chat for an app and the user explicitly approves that before the actual app building starts.

that's a pretty involved change (and I still need to figure out what the right UX is) so I'd recommend punting on this feature request for now.

Issue #1200, opened 2025-09-05
Allow chat UI to be on the right side


wwwillchen:
From customer: 

> Please consider allowing the AI conversation window to be positioned on the right, not just the left.  Similar to how you can deploy it in Zed or Windsurf.

tharani-2006:
Shall i work on this issue?

wwwillchen:
we've decided to not support this due to UX complexity.

PR #1342, opened 2025-09-21
implementing switch chat side feature

azizmejri1:
This PR implements switch chat side feature mentioned in issue #1200 
    

---

## Summary by cubic
Adds a toggle to switch the chat panel between the left and right side of the page. Implements the “switch chat side” feature from issue #1200.

- **New Features**
  - Added a header button to move the chat panel left/right (ArrowLeftFromLine/ArrowRightFromLine).
  - Reorders panels using Tailwind order classes so layout flips without remounting content.
  - Manages isChatOnLeft state in chat.tsx and wires it through ChatPanel and ChatHeader props.

### DIFF
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -13,12 +13,16 @@ interface ChatPanelProps {
   chatId?: number;
   isPreviewOpen: boolean;
   onTogglePreview: () => void;
+  onSwitchChatSide: () => void;
+  isChatOnLeft: boolean;
 }
 
 export function ChatPanel({
   chatId,
   isPreviewOpen,
   onTogglePreview,
+  onSwitchChatSide,
+  isChatOnLeft,
 }: ChatPanelProps) {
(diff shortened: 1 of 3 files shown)


wwwillchen:
thanks for the PR. it definitely works, but it feels kind of funky UX-wise because the panel header buttons basically end up on the wrong side:

image

i know I filed #1200 but I don't feel totally sold on it, it feels like it adds complexity for IMO not a ton of benefit.  if you're open, maybe I could recommend some other issues that you could work on?

azizmejri1:
Yeah , i'm definetly intersted in that . I'd love to hear your recommendation about other issues to tackle.

wwwillchen:
@azizmejri1 i've created a list of issues that I think would be good to work on: https://github.com/dyad-sh/dyad/issues?q=state%3Aopen%20label%3A%22good%20issues%20for%20contributors%22 please feel free to leave a comment if something is unclear / email me. thanks!

PR #1359, opened 2025-09-23
feat: text-to-speech for AI responses

princeaden1:
# feat: text-to-speech for AI responses

Add text-to-speech functionality to AI chat responses with e2e testing.

## Changes
- **TTS Button**: Play/stop button for AI messages with state management
- **Voice Support**: Uses browser's Web Speech API with voice selection
- **UI States**: Visual feedback with play/stop icons and tooltips
- **E2E Tests**: Complete test coverage following existing patterns

https://github.com/user-attachments/assets/3891652d-2295-47eb-9de8-3c34d3f1b164

    

---

## Summary by cubic
Adds text-to-speech to assistant messages so users can listen to chat replies. Includes a play/stop control and end-to-end tests.

- New Features
  - Play/stop button on assistant messages with icons and tooltips
  - Web Speech API integration with voice selection and support check
  - Text cleaning strips markdown, HTML, and Dyad tags before reading
  - E2E test mocks speechSynthesis and verifies state changes

### DIFF
--- a/e2e-tests/text_to_speech.spec.ts
+++ b/e2e-tests/text_to_speech.spec.ts
@@ -3,7 +3,6 @@ import { expect } from "@playwright/test";
 
 test("text to speech - basic functionality", async ({ po }) => {
   await po.setUp({ autoApprove: true });
-  await po.importApp("minimal");
 
   // Mock speechSynthesis API
   await po.page.addInitScript(() => {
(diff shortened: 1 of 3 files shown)


wwwillchen:
thanks @princeaden1. i pulled it down and it does work, but I'm not super sure about the use case - is it for accessibility or some other reason? in general - i'd recommend opening an issue for new feature requests to discuss the use case and see whether it's aligned with the projects direction. for example, right now it's reading the content in the think tags which can be quite verbose and somewhat confusing for users IMO.

princeaden1:
@wwwillchen Thanks for taking time to review the PR🙏. This PR is for both accessibility and to let devs listen to responses if they don’t want to read through chats (similar to ChatGPT’s feature). The think tags can be also removed from the spoken output. I'll also make sure to open an issue first for future feature requests to be first discussed.

PR #1362, opened 2025-09-24
Arquivo Dockerfile

ZumMkt:
## Summary by cubic
Add a multi-stage Dockerfile using Node 20-alpine to containerize the app. It installs dependencies with Yarn (fallback to npm), optionally runs the build, exposes port 3000, and starts with yarn start.


wwwillchen:
hi @JrZum thanks for opening a PR. we ask that all code & comments be done in english to make it easier for collaboration. since this seems like something more tailored for your individual needs, i'll go ahead and close this

PR #1392, opened 2025-09-29
Update Gemini 2.5 Flash to point to latest

princeaden1:


wwwillchen:
@princeaden1 i haven't been able to get good results with flash-lite - it seems to not be able to follow dyad's system prompt. have you gotten good results with it?

based on https://developers.googleblog.com/en/continuing-to-bring-you-our-latest-models-with-an-improved-gemini-2-5-flash-and-flash-lite-release/ - i think we should update flash 2.5 to instead use `gemini-flash-latest`

princeaden1:
@wwwillchen thanks! Although Flash-lite works for me in some cases, I've also noticed that the results don’t flow well, just like you mentioned. I’ll go ahead and update the PR to use the latest version of Flash instead.

wwwillchen:
thanks!

Issue #1416, opened 2025-10-01
[bug] Can't scroll chat while thinking


Coder-Jake:
## Bug Description (required)
The chat feed keeps snapping back to the bottom while it's thinking, so I can't read smoothly until it's finished.

## Steps to Reproduce (required)
Try scrolling through chat while it's processing a request

## Expected Behavior (required)
No snapping. Maybe have a "scroll to bottom" icon pop-up when the thinking is complete: Click once to scroll to the top of the thinking (to follow it's thoughts), click twice to scroll to bottom (to hit approve)

## Actual Behavior (required)
"chat feed keeps snapping back to the bottom while it's thinking"

Also this app is amazing! Thank you so much!!!

## System Information
- Dyad Version: 0.22.0
- Platform: darwin
- Architecture: arm64
- Node Version: v22.14.0
- PNPM Version: n/a
- Node Path: /usr/local/bin/node
- Telemetry ID: f203db58-7f8e-448e-9a7d-eb79dbb7ef0a
- Model: auto:auto | customId: undefined

## Logs
```
Run `npm audit` for details.

> vite_react_shadcn_ts@0.0.0 dev
> vite --port 32100


  VITE v5.4.10  ready in 182 ms


  ➜  Network: http://192.168.88.25:32100/

[2025-10-01 15:15:24.307] [error] (app_handlers) App 3 (PID: 3668) stderr: Browserslist: browsers data (caniuse-lite) is 12 months old. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme

[2025-10-01 15:16:36.797] [error] (pro_handlers) LLM Gateway API key (Dyad Pro) is not configured.

[2025-10-01 15:18:29.175] [warn]  (runShellCommand) Stderr from "pnpm --version": /bin/sh: pnpm: command not found
[2025-10-01 15:18:29.176] [error] (runShellCommand) Command "pnpm --version" failed with code 127

```

wwwillchen:
thanks for the feedback - agree the auto-scrolling behavior is funky, if you scroll fast enough away from the bottom it stops auto-scrolling, but it's quite glitchy.

rakibulrocky14:
Problem solved!!

wwwillchen:
should be fixed in next release

PR #1425, opened 2025-10-01
Enhance ChatPanel with scroll-to-bottom button and improve scrolling

rakibulrocky14:
# **Fix Auto-Scrolling Behavior in Chat**

This update addresses the issue of glitchy auto-scrolling that hindered users from reading messages during streaming. It introduces a smart scroll-to-bottom button with a two-click functionality. The changes include:

- Improved scroll detection that tracks the distance from the bottom instead of the direction of scrolling.
- Reduced the auto-scroll threshold from 280 pixels to 100 pixels for less aggressive scrolling.
- Added a floating scroll-to-bottom button that appears when the user has scrolled more than 150 pixels away from the bottom.
  - The first click will scroll to the latest message, and the second click will scroll all the way to the bottom.
- The button has been styled to align with the Dyad design system, featuring a glassmorphism effect.
- Increased the user scroll timeout from 1 second to 2 seconds for enhanced user experience.

**Files Modified:**
- `src/components/ChatPanel.tsx` - Updated scroll logic, button UI, and event handlers 
- `src/components/chat/MessagesList.tsx` - Fixed positioning for button overlay.

Issues: #1416 

# Auto Generated by Copilot
This pull request introduces a scroll-to-bottom button for the chat panel, improving usability for users who have scrolled away from the latest messages. The button appears when the user is not near the bottom of the message list and provides a two-step scroll: first to the latest message, then to the very bottom. Additional refinements to scroll detection and auto-scrolling behavior help ensure a smoother chat experience.

**Scroll-to-bottom button feature:**
* Added a floating scroll-to-bottom button to the chat panel that appears when the user is scrolled away from the latest messages. The button enables a two-step scroll: first to the latest message, then to the absolute bottom. (`src/components/ChatPanel.tsx`, `src/components/chat/MessagesList.tsx`) [[1]](diffhunk://#diff-a1f481e60c6695a8151070b5443cb917bd0a9cf522f8b2a21b538fe290cae2b2R29-R116) [[2]](diffhunk://#diff-a1f481e60c6695a8151070b5443cb917bd0a9cf522f8b2a21b538fe290cae2b2R174-R200) [[3]](diffhunk://#diff-9eadeffc0e0b9a2884e1a1ef783476627eb3bcb83812ae428bc764ee28dd84d9L57-R57)

**Scroll detection and auto-scroll improvements:**
* Refined logic to detect when the user is near the bottom of the message list and show or hide the scroll button accordingly. (`src/components/ChatPanel.tsx`)
* Improved auto-scrolling behavior to only scroll when the user is close to the bottom and when new messages arrive during streaming. (`src/components/ChatPanel.tsx`)
* Increased timeout for resetting user scroll state to reduce flickering of the scroll button. (`src/components/ChatPanel.tsx`)
* Updated the message list container to use absolute positioning for proper overlay of the scroll button. (`src/components/chat/MessagesList.tsx`)
    

---

## Summary by cubic
Fixes glitchy auto-scrolling during streaming and adds a two-step scroll-to-bottom button for faster navigation without losing your place. Addresses #1416.

- **New Features**
  - Floating scroll-to-bottom button appears when >150px from bottom.
  - Two-click flow: first to latest message, second to the absolute bottom.
  - Message list uses absolute positioning so the button overlays correctly.

- **Bug Fixes**
  - Auto-scroll now based on distance from bottom; only triggers within 100px.
  - Increased user scroll timeout to 2s to reduce flicker and reset button state near bottom.
  - Auto-scroll during streaming runs only when the user isn’t actively scrolling.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
 npx lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,1 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
 npx lint-staged
--- a/src/components/ChatPanel.tsx
+++ b/src/components/ChatPanel.tsx
@@ -1,6 +1,6 @@
 import { useState, useRef, useEffect, useCallback } from "react";
 import { useAtom, useAtomValue } from "jotai";
-import { chatMessagesAtom, chatStreamCountAtom } from "../atoms/chatAtoms";
+import { chatMessagesAtom, chatStreamCountAtom, isStreamingAtom } from "../atoms/chatAtoms";
 import { IpcClient } from "@/ipc/ipc_client";
 
 import { ChatHeader } from "./chat/ChatHeader";
@@ -8,6 +8,8 @@ import { MessagesList } from "./chat/MessagesList";
 import { ChatInput } from "./chat/ChatInput";
 import { VersionPane } from "./chat/VersionPane";
 import { ChatError } from "./chat/ChatError";
+import { Button } from "@/components/ui/button";
+import { ArrowDown } from "lucide-react";
 
 interface ChatPanelProps {
   chatId?: number;
@@ -24,6 +26,7 @@ export function ChatPanel({
   const [isVersionPaneOpen, setIsVersionPaneOpen] = useState(false);
   const [error, setError] = useState<string | null>(null);
   const streamCount = useAtomValue(chatStreamCountAtom);
+  const isStreaming = useAtomValue(isStreamingAtom);
   // Reference to store the processed prompt so we don't submit it twice
 
   const messagesEndRef = useRef<HTMLDivElement | null>(null);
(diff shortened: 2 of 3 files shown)


rakibulrocky14:
fixed the prettier issue

.husky/pre-commit
```diff
@@ -1,4 +1 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
 npx lint-staged
```
wwwillchen, line 4: let's revert this change

wwwillchen (changes requested):
thanks for the PR!
I like the scroll to bottom button, but the positioning is a little weird - I think it should be bottom-center in the chat pane area (instead of bottom right): 

Image

also, the intended behavior is that if you're near the bottom, it will stick to bottom, however when I use a model like gemini 2.5 flash which generates pretty quickly, it's not sticking to the bottom properly.

.husky/pre-commit
rakibulrocky14: ok

rakibulrocky14:
_@wwwillchen can you check now? I hope it's better now_

wwwillchen (changes requested):
The stick to bottom scroll behavior is still not working as expected. if too much AI content is generated at once, it's not sticking to bottom.

I think for now, let's just add the "scroll-to-bottom" button and then keep the existing stick-to-bottom scroll behavior (as this new behavior feels a bit worse and a regression). Thanks.

rakibulrocky14:
@wwwillchen is it good now?

wwwillchen:
@rakibulrocky14 i pulled it down and the auto-scroll behavior still isn't right, i've pulled your PR and made some changes to simplify it in #1484 - I'll go ahead and close this pr, thanks

PR #1432, opened 2025-10-02
fix: show dyad-command buttons within assistant responses instead of hiding tags (#1421)

### LINKED ISSUE #1421: [bug] <dyad-command> tags are not being displayed which is confusing
Image

## Bug Description (required)
Please try to  again message is not clear

Image

## System Information
- Dyad Version: 0.22.0
- Platform: win32
- Architecture: x64
- Node Version: v20.18.1
- PNPM Version: 10.13.1
- Node Path: C:\nvm4w\nodejs\node.exe
C:\Program Files\nodejs\node.exe
- Telemetry ID: 84640c25-9e2c-4815-84ab-6cfc02e50f65
- Model: auto:auto | customId: undefined

## Logs
```
[2025-10-01 13:09:37.396] [error] (pro_handlers) Failed to fetch user budget. Status: 401. Body: {"error":{"message":"Authentication Error, Invalid proxy server token passed. Received API Key = sk-...y2qQ, Key Hash (Token) =194559cb7453a77b0070a4ac01931b604754971c50d6a78713599452b2c2af54. Unable to find token in cache or `LiteLLM_VerificationTokenTable`","type":"token_not_found_in_db","param":"key","code":"401"}}
[2025-10-01 13:09:39.585] [error] (tsc) TSC worker exited with code 1 for app C:\Users\jnass\dyad-apps\KeyDrop
C:\Program Files\nodejs\node.exe

```

princeaden1:
### Problem
Users were confused because `<dyad-command>` tags were being hidden, making it unclear what actions were available.

### Solution
Render dyad-command tags as visible buttons, so users can see and interact with available commands (e.g., "restart").

Closes #1421
    

---

## Summary by cubic
Show dyad-command tags as visible buttons in assistant responses so users can see and trigger actions (e.g., restart). Closes #1421.

- **Bug Fixes**
  - Render dyad-command tags to buttons in DyadMarkdownParser.
  - Export mapActionToButton from ChatInput for reuse.
  - Support id/action/path attributes; safely no-op if parsing fails.

### DIFF
--- a/src/components/chat/DyadMarkdownParser.tsx
+++ b/src/components/chat/DyadMarkdownParser.tsx
@@ -500,19 +500,13 @@ function renderCustomTag(
       return null;
 
     case "dyad-command":
-      try {
-        let action;
-        if (attributes.type) {
-          action = {
-            id: attributes.type,
-          } as SuggestedAction;
-        } else {
-          return null;
-        }
+      if (attributes.type) {
+        const action = {
+          id: attributes.type,
+        } as SuggestedAction;
         return <>{mapActionToButton(action)}</>;
-      } catch {
-        return null;
       }
+      return null;
 
     default:
       return null;
--- a/src/components/chat/DyadMarkdownParser.tsx
+++ b/src/components/chat/DyadMarkdownParser.tsx
@@ -502,25 +502,14 @@ function renderCustomTag(
     case "dyad-command":
       try {
         let action;
-        if (attributes.id || attributes.action) {
+        if (attributes.type) {
           action = {
-            id: attributes.id || attributes.action,
-            path: attributes.path,
-          } as SuggestedAction;
-        } else if (Object.keys(attributes).length > 0) {
-          const firstKey = Object.keys(attributes)[0];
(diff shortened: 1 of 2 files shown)


src/components/chat/DyadMarkdownParser.tsx
```diff
@@ -498,8 +500,30 @@ function renderCustomTag(
       return null;
 
     case "dyad-command":
-      // Don't render anything for dyad-command
-      return null;
+      try {
+        let action;
+        if (attributes.id || attributes.action) {
+          action = {
+            id: attributes.id || attributes.action,
```
wwwillchen: this feels a bit too indirect (even though it does work). I think you can use attributes.type and map it to `id` and that should work for all the existing cases.

wwwillchen (changes requested):
thanks! left a code suggestion but UX-wise, it's good :)

princeaden1:
@wwwillchen the code has been simplified to be more direct.

wwwillchen (approved):
thanks!

PR #1454, opened 2025-10-06
GitHub Import Feature: Import repositories/projects from GitHub (#1424)

### LINKED ISSUE #1424: Clone project from GitHub

princeaden1:
## Summary
Adds the ability to import GitHub repositories directly into Dyad from the home screen, complementing the existing local folder import feature.
- GitHub Import Modal: New modal accessible from home screen via "Import from Github" button with two Import methods
- Select project from GitHub repositories list
- Clone from any GitHub URL
- Advanced Options: Optional custom install/start commands (defaults to project's package.json scripts)
- Auto AI_RULES Generation: Automatically generates AI_RULES.md if not present in imported repo

closes #1424
    

---

## Summary by cubic
Adds a GitHub import flow from the home screen so users can clone repos via their list or any URL, with optional install/start commands and automatic AI_RULES.md generation. Addresses Linear #1424 by enabling seamless project setup from GitHub.

- **New Features**
  - Import modal with two tabs: Your Repositories and From URL.
  - Advanced options for install/start commands with validation; defaults used when both are empty.
  - After cloning, navigate to chat and auto-generate AI_RULES.md if missing.
  - New IPC handler github:clone-repo-from-url with token auth support, plus IpcClient method and preload channel.
  - E2E tests cover modal open, auth, import via URL/repo list, and advanced options.

- **Dependencies**
  - Added @radix-ui/react-tabs for the modal tab UI.


src/components/chat/GithubRepoModal.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+  useEffect(() => {
+    if (isOpen) {
+      if (isAuthenticated) {
+        fetchRepos();
+      }
+      // Reset state when modal opens
+      setUrl("");
+      setError(null);
+      setInstallCommand("pnpm install");
```
wwwillchen: This is actually a mistake in the existing import dialog (if you can fix it, that would be great!), by default we should **not** set the install or start command. This is because when the install or start command are unset, we have a default command (see https://github.com/dyad-sh/dyad/blob/d83f5fec8366b48eca2a440ddc7168b1ea1c7c9a/src/ipc/handlers/app_handlers.ts#L1488)

The placeholder value should still say "pnpm install" and "pnpm dev" but we shouldn't preset these form values. Not every user pnpm installed, and our default command tries pnpm and fallsback to npm https://github.com/dyad-sh/dyad/blob/d83f5fec8366b48eca2a440ddc7168b1ea1c7c9a/src/ipc/handlers/app_handlers.ts#L56

src/components/chat/GithubRepoModal.tsx
```diff
@@ -0,0 +1,346 @@
...
+        </Alert>
+
+        <Tabs defaultValue="your-repos" className="w-full">
+          <TabsList className="grid w-full grid-cols-2">
+            <TabsTrigger value="your-repos">Your Repositories</TabsTrigger>
+            <TabsTrigger value="from-url">From URL</TabsTrigger>
+          </TabsList>
+
+          <TabsContent value="your-repos" className="space-y-4">
+            {loading && (
```
wwwillchen: if you've already loaded the repo before (e.g. opened dialog, close dialog and re-open dialog), it's a bit jarring to have a loading spinner that then closes. Can we instead put the spinner next to the tab header? What do you think?

src/components/chat/GithubRepoModal.tsx
wwwillchen, line 1: For this dialog and the existing dialog, when the Dyad app has a short height, the dialog gets cutoff. We should cap dialog height using something like `max-h-[80vh]` (tailwind class)

Image

src/ipc/handlers/github_handlers.ts
```diff
@@ -627,6 +627,109 @@ async function handleDisconnectGithubRepo(
...
+    url,
+    installCommand,
+    startCommand,
+  }: {
+    url: string;
+    installCommand?: string;
+    startCommand?: string;
+  },
+): Promise<{
+  success: boolean;
```
wwwillchen: i'd probably avoid doing this `success: boolean` for future code - i know some of the code does this pattern already, but i've found it really easy to forget to handle the failure case and i think throwing an error/exception directly is more idiomatic and easier to handle.

src/ipc/handlers/github_handlers.ts
```diff
@@ -627,6 +627,109 @@ async function handleDisconnectGithubRepo(
...
+  app?: typeof schema.apps.$inferSelect;
```
wwwillchen: I think you can just use ipc_types.ts App interface

src/components/chat/GithubRepoModal.tsx
```diff
@@ -0,0 +1,346 @@
...
+      const fetchedRepos = await IpcClient.getInstance().listGithubRepos();
+      setRepos(fetchedRepos);
+    } catch (err: any) {
+      setError(err.message || "Failed to fetch repositories.");
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  const checkAndGenerateAiRules = async (appId: number, appPath: string) => {
```
wwwillchen: i'm noticing even when i import an app **with** AI_RULES.md it's still trying to generate this file even though it already exists. could you take a look?

wwwillchen (changes requested):
Thanks! overall, good work!

A few higher-level comments about the feature and apologies for not providing more details upfront.

**UX-wise**

- We should have a single button and dialog for importing apps, whether it's locally or via GitHub. From a user perspective, it's a similar flow and intent. We should use the existing import button and then also add a GitHub logo to it, to make it clearer you can also import from GitHub.
- Let's have three tabs for the import dialog, whose order roughly correspond to what I think is the most common use cases:
  - GitHub URL (what you're adding)
  - Your GitHub Repos (what you're adding)
  - Local (the current import app dialog)
- For GitHub URL tab, we should not gate this behind whether the user has authenticated to GitHub because sometimes users may want to import a public GitHub URL and it's not necessary to authenticate w/ GitHub.
- Similar to the existing import dialog, we should allow users to specify an alternate app name, this way if there's a naming conflict, they can pick something else (both for the GitHub URL and repo list picker)

I've left a few more comments inline

princeaden1:
@wwwillchen, thanks for the detailed feedback and clarifications! I'll push an update soon to reflect these changes.

princeaden1:
@wwwillchen I've fixed the issue with generating AI_rules.md for repositories that already have the file. I've also resolved the issue with the default pnpm configuration for file import.

src/components/ImportAppDialog.tsx
```diff
@@ -173,158 +294,366 @@ export function ImportAppDialog({ isOpen, onClose }: ImportAppDialogProps) {
...
-              disabled={selectFolderMutation.isPending}
-              className="w-full"
-            >
-              {selectFolderMutation.isPending ? (
-                <Loader2 className="mr-2 h-4 w-4 animate-spin" />
+        <Tabs defaultValue="local-folder" className="w-full">
+          <TabsList className="grid w-full grid-cols-3">
+            <TabsTrigger value="local-folder">Local Folder</TabsTrigger>
+            <TabsTrigger value="github-repos">Your Repositories</TabsTrigger>
+            <TabsTrigger value="github-url">From URL</TabsTrigger>
```
wwwillchen: I would call this: "GitHub URL" (I think the "From" is unnecessary and implied)

src/components/ImportAppDialog.tsx
```diff
@@ -173,158 +294,366 @@ export function ImportAppDialog({ isOpen, onClose }: ImportAppDialogProps) {
...
-              onClick={handleSelectFolder}
-              disabled={selectFolderMutation.isPending}
-              className="w-full"
-            >
-              {selectFolderMutation.isPending ? (
-                <Loader2 className="mr-2 h-4 w-4 animate-spin" />
+        <Tabs defaultValue="local-folder" className="w-full">
+          <TabsList className="grid w-full grid-cols-3">
+            <TabsTrigger value="local-folder">Local Folder</TabsTrigger>
+            <TabsTrigger value="github-repos">Your Repositories</TabsTrigger>
```
wwwillchen: Let's call this "Your GitHub Repos" (make it clearer this is GitHub)

src/components/ImportAppDialog.tsx
```diff
@@ -173,158 +294,366 @@ export function ImportAppDialog({ isOpen, onClose }: ImportAppDialogProps) {
...
+                        </AccordionContent>
+                      </AccordionItem>
+                    </Accordion>
+                  </>
+                )}
+              </>
+            )}
+          </TabsContent>
+          <TabsContent value="github-url" className="space-y-4">
+            <div className="space-y-2">
```
wwwillchen, line 598: I think this should be below the GitHub URL (ask for optional parameters  after required parameters)

And then once the user inputs the GitHub URL, we should prefill the app name and run the app name check so it shows a nice error message inline (right now you get the error toast)

src/ipc/handlers/github_handlers.ts
```diff
@@ -627,6 +629,121 @@ async function handleDisconnectGithubRepo(
     })
     .where(eq(apps.id, appId));
 }
+// --- GitHub Clone Repo from URL Handler ---
+async function handleCloneRepoFromUrl(
+  event: IpcMainInvokeEvent,
+  params: CloneRepoParams,
+): Promise<
```
wwwillchen: Similar to CloneRepoParams, I would create an IPC type called CloneRepoReturnType and then use it here and in ipc_client.ts, this way it won't get out of sync in the future.

wwwillchen (changes requested):
Overall, this is good! I've left a few comments to improve the UX and code maintainability a bit. once that's resolved, we can merge. thanks!

wwwillchen (approved):
looks good, thanks!

PR #1500, opened 2025-10-10
Prompt enhance

orev7s:
a prompt enchaner button + toggle

image

added a prompt enhancer that uses the model you have setup to enhance prompts. it can be toggled on for auto enhance or just clicked when needed if you switch it to "button mode" in settings. 
    

---

## Summary by cubic
Added prompt enhancement to chat inputs with a button or auto-enhance toggle, using your configured model to improve prompts before sending. Also added “Your templates” in the Hub: import a local folder, manage templates, and use them when creating apps.

- **New Features**
  - Prompt Enhancement: choose Button (Enhance+Send) or Toggle in Settings; optional auto-enhance default.
  - Works in Home and Chat inputs; uses the selected model (test builds return “[enhanced] <text>”).
  - Your templates: import a project folder via a dialog; templates are listed under “Your templates”.
  - Delete user templates with confirmation; app creation copies from the selected user template folder.
  - E2E coverage for Enhance button mode and add/delete user templates.

### DIFF
--- a/e2e-tests/enhance_button.spec.ts
+++ b/e2e-tests/enhance_button.spec.ts
@@ -1,0 +1,31 @@
+import { test, expect } from "@playwright/test";
+
+// This test suite assumes IS_TEST_BUILD so prompt enhancement returns "[enhanced] <text>"
+
+test.describe("Prompt enhancement - button mode", () => {
+  test("Home: Enhance button enhances and starts chat", async ({ page }) => {
+    // Go to settings
+    await page.goto("/");
+    await page.getByRole("link", { name: "Settings" }).click();
+
+    // Navigate to AI Settings -> Prompt Enhancement and choose Button
+    await page.locator("text=AI Settings").scrollIntoViewIfNeeded();
+    await page.locator("text=Prompt Enhancement").scrollIntoViewIfNeeded();
+    await page.getByRole("button", { name: "Button (Enhance+Send)" }).click();
+
+    // Go home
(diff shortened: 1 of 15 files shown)


wwwillchen (commented):
Thanks for the PR. A few high-level feedback comments:

1. Let's keep the PRs focused. This includes the template stuff from the previous PR which makes it hard to review.
2. I think prompt enhance should just be a button and **not** a toggle. I generally don't think it's the right UX to enhance a prompt every time, especially without a human being able to review it. UX-wise, it should just be an icon button like a lightning bolt - I think bolt.new has a good UX for this feature
3. there needs to be a loading UX while the prompt enhance is running (since it can take several seconds), i think having a loading spinner in the input box would be good enough.

orev7s:
> Thanks for the PR. A few high-level feedback comments:
> 
> 1. Let's keep the PRs focused. This includes the template stuff from the previous PR which makes it hard to review.
> 2. I think prompt enhance should just be a button and **not** a toggle. I generally don't think it's the right UX to enhance a prompt every time, especially without a human being able to review it. UX-wise, it should just be an icon button like a lightning bolt - I think bolt.new has a good UX for this feature
> 3. there needs to be a loading UX while the prompt enhance is running (since it can take several seconds), i think having a loading spinner in the input box would be good enough.

oh man i didnt notice i complicated stuff together. ill rework this PR and open another branch same for the custom template one okay? you can close this.

PR #1577, opened 2025-10-19
feat: allow manual Node.js path configuration

### LINKED ISSUE #1050: Allow manually configuring node.js / pnpm path
See: https://github.com/dyad-sh/dyad/issues/990
This comes up quite often

princeaden1:
Add ability to manually configure Node.js path for users who have Node.js
installed but not in their system PATH.

Features:
- Browse and select custom Node.js installation folder
- Visual status indicator showing Node.js version or "Not found"
- Reset to system default PATH option
- Manual configuration option in setup banner
- Real-time Node.js status checking

closes #1050

    

---

## Summary by cubic
Adds manual Node.js path configuration so the app works even when Node isn’t on PATH, fulfilling #1050. Users can browse to their install, reset to default, and see real-time status in Settings and during setup.

- New Features
  - Settings: NodePathSelector to browse a Node.js folder, show version/“Not found” status, and reset to system PATH (persists customNodePath).
  - Setup banner: manual config flow with a folder picker if Node is already installed.
  - IPC: select-node-folder, set-node-path, get-node-path; reloads env and prepends custom path to PATH.
  - Real-time Node.js status check with visual indicator (CheckCircle on valid).
  - E2E tests for browse, reset, and valid-status display.

### DIFF
--- a/src/components/NodePathSelector.tsx
+++ b/src/components/NodePathSelector.tsx
@@ -24,17 +24,18 @@ export function NodePathSelector() {
     checkNodeStatus();
   }, [settings?.customNodePath]);
 
+  const fetchSystemPath = async () => {
+    try {
+      const debugInfo = await IpcClient.getInstance().getSystemDebugInfo();
+      setSystemPath(debugInfo.nodePath || "System PATH (not available)");
+    } catch (err) {
+      console.error("Failed to fetch system path:", err);
+      setSystemPath("System PATH (not available)");
+    }
+  };
+
   useEffect(() => {
     // Fetch system path on mount
-    const fetchSystemPath = async () => {
-      try {
-        const debugInfo = await IpcClient.getInstance().getSystemDebugInfo();
-        setSystemPath(debugInfo.nodePath || "System PATH (not available)");
-      } catch (err) {
-        console.error("Failed to fetch system path:", err);
-        setSystemPath("System PATH (not available)");
-      }
-    };
     fetchSystemPath();
   }, []);
 
@@ -68,8 +69,10 @@ export function NodePathSelector() {
         // Recheck Node.js status
         await checkNodeStatus();
         showSuccess("Node.js path updated successfully");
-      } else if (result.path === null) {
-        showError("Invalid Node.js path selected");
+      } else if (result.path === null && result.canceled === false) {
+        showError(
+          `Could not find Node.js at the path "${result.selectedPath}"`,
+        );
       }
     } catch (error: any) {
       showError(`Failed to set Node.js path: ${error.message}`);
@@ -84,6 +87,7 @@ export function NodePathSelector() {
       // Reload environment to use system PATH
       await IpcClient.getInstance().reloadEnvPath();
       // Recheck Node.js status
+      await fetchSystemPath();
       await checkNodeStatus();
       showSuccess("Reset to system Node.js path");
     } catch (error: any) {
--- a/src/components/NodePathSelector.tsx
+++ b/src/components/NodePathSelector.tsx
@@ -64,7 +64,6 @@ export function NodePathSelector() {
         // Save the custom path to settings
         await updateSettings({ customNodePath: result.path });
         // Update the environment PATH
-        await IpcClient.getInstance().setNodePath(result.path);
         await IpcClient.getInstance().reloadEnvPath();
         // Recheck Node.js status
         await checkNodeStatus();
--- a/src/components/NodePathSelector.tsx
+++ b/src/components/NodePathSelector.tsx
@@ -1,0 +1,183 @@
+import { useState, useEffect } from "react";
+import { Label } from "@/components/ui/label";
+import { Button } from "@/components/ui/button";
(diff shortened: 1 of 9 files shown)


src/components/NodePathSelector.tsx
```diff
@@ -0,0 +1,183 @@
...
+        // Save the custom path to settings
+        await updateSettings({ customNodePath: result.path });
+        // Update the environment PATH
+        await IpcClient.getInstance().setNodePath(result.path);
+        await IpcClient.getInstance().reloadEnvPath();
+        // Recheck Node.js status
+        await checkNodeStatus();
+        showSuccess("Node.js path updated successfully");
+      } else if (result.path === null) {
+        showError("Invalid Node.js path selected");
```
wwwillchen: I'd make this a little more descriptive and say that "Could not find Node.js at the path /foo/path" - this makes it easier to understand

src/components/SetupBanner.tsx
```diff
@@ -223,6 +249,42 @@ export function SetupBanner() {
...
+                      onClick={() => setShowManualConfig(!showManualConfig)}
+                      className="text-sm text-blue-600 dark:text-blue-400 hover:underline"
+                    >
+                      Node.js already installed? Configure path manually →
+                    </button>
+
+                    {showManualConfig && (
+                      <div className="mt-3 p-3 bg-gray-50 dark:bg-gray-800 rounded-lg">
+                        <p className="text-sm text-gray-700 dark:text-gray-300 mb-3">
+                          Select the folder where Node.js is installed (usually
```
wwwillchen: I think I would just remove this parenthetical comment, since I think people who _already_ have Node.js installed probably know this and if they don't, we can make it clearer in the error message (e.g. did not find "node.exe" at x/y/z path) - see other comment

src/ipc/handlers/node_handlers.ts
```diff
@@ -1,14 +1,18 @@
...
 import { runShellCommand } from "../utils/runShellCommand";
 import log from "electron-log";
+import { existsSync } from "fs";
+import { join } from "path";
+import { writeSettings, readSettings } from "../../main/settings";
 
 const logger = log.scope("node_handlers");
 
 export function registerNodeHandlers() {
+  const settings = readSettings();
```
wwwillchen: do the read at L59 not up here. this will only execute once (when the dyad app starts up) and give you stale data.

src/preload.ts
```diff
@@ -136,6 +136,9 @@ const validInvokeChannels = [
   // We can't detect with IS_TEST_BUILD in the preload script because
   // it's a separate process from the main process.
   "supabase:fake-connect-and-set-project",
+  "select-node-folder",
```
wwwillchen: put this next to the other node-related IPCs (L53)

src/ipc/handlers/node_handlers.ts
```diff
@@ -52,6 +56,62 @@ export function registerNodeHandlers() {
...
+    if (platform() === "win32") {
+      process.env.PATH = `${nodePath};${process.env.PATH}`;
+    } else {
+      process.env.PATH = `${nodePath}:${process.env.PATH}`;
+    }
+
+    logger.info("Custom Node.js path set:", nodePath);
+  });
+
+  ipcMain.handle("get-node-path", async () => {
```
wwwillchen: remove this IPC - looks like it's not needed

src/components/NodePathSelector.tsx
```diff
@@ -0,0 +1,183 @@
...
+      setIsCheckingNode(false);
+    }
+  };
+  const handleSelectNodePath = async () => {
+    setIsSelectingPath(true);
+    try {
+      // Call the IPC method to select folder
+      const result = await IpcClient.getInstance().selectNodeFolder();
+      if (result.path) {
+        // Save the custom path to settings
```
wwwillchen, line 64: looking at the code, I don't think setNodePath IPC is actually needed.
you're updating the settings via updateSettings and then reloadEnvPath is updating the path.

wwwillchen (changes requested):
Thanks for the PR, I definitely think this will help our users. Functionality-wise, this works, but I think the code here can be cleaner (see comments below).

- I noticed if I open the browse for node.js file picker and click close/cancel, it will show an error toast, we shouldn't show an error in this case

On a side note:
- I noticed an edge case but I don't think it's worth fixing: essentially if you set a custom node.js path and then reset to default, because the PATH has the custom node.js path, it will show the custom node.js path as coming from the PATH until you restart Dyad. I feel like this is a pretty obscure edge case and the fix IMO is not worth it.

src/components/NodePathSelector.tsx
wwwillchen: @princeaden1 could you address this comment? thanks

princeaden1: @wwwillchen Thanks, I've gone through the code again and removed setNodePath.

wwwillchen (approved):
thanks!

Issue #251, opened 2025-05-26
Toggle Between Screen Sizes


beaugast:
Ability to toggle the screen size in the preview. Like mobile view, desktop view, tablet view, etc.

wwwillchen:
Because the preview size is currently contained to the right side panel, there's not a lot of space to show the desktop or tablet view. We could do one of two things:

1. Zoom out the preview (e.g. for desktop view, it would probably be very zoomed-out)
2. Pop open the preview in a full-screen dialog and then you can toggle between mobile/tablet/desktop, etc.

What do you think? Are you typically working on a big monitor?

Also, not sure if you're aware, but Chrome Developer Tools has a device mode that lets you do this (if you're using preview in a new window already): https://developer.chrome.com/docs/devtools/device-mode but I do think it would be nice to do something natively in Dyad.

beaugast:
Personally, I like option 1. We should keep the current view when we aren't looking at these custom views since it is pretty great as is.

PR #1582, opened 2025-10-19
feat: allow toggling between screen sizes

azizmejri1:
Implement the feature requested in issue #251 that allows users to toggle between screen sizes
    

---

## Summary by cubic
Adds a screen size toggle to the preview panel with device presets, custom dimensions, rotation, and smart scaling. Presets are persisted via a new device_presets table and can be managed in-app. Addresses #251.

- **New Features**
  - Screen size menu: select a device preset, set width/height in px, rotate dimensions, and auto-scale the iframe with a zoom indicator; closing resets to 100%.
  - Manage presets dialog: view defaults, add/edit/delete custom presets (defaults are read-only) with success/error toasts.
  - Persistence: new SQLite device_presets table, seeded with common devices on startup; IPC handlers and preload channels for get/add/update/delete.
  - Tests: Playwright e2e for menu toggle, preset selection, custom input, rotation, and presets dialog.


wwwillchen:
thanks for the PR @azizmejri1 - I think we can simplify this feature both functionality & implementation-wise - right now this feels like Chrome DevTools device mode which feels like an overkill for most of dyad's user use cases (apologies if the issue was ambiguous). I think the device mode button should open a [popover](https://ui.shadcn.com/docs/components/popover) which contains a [button group](https://ui.shadcn.com/docs/components/button-group) of three icon buttons: desktop, tablet, and mobile.

wwwillchen (changes requested):
thanks for the updates! this is definitely an improvement

a few thoughts on the user experience:
- I don't think we should zoom out to fit the screen. Instead, we should just cap the width so that it does not **exceed** the selected device mode width. Check out what Lovable does for their responsive mode.
- I don't think we need "Original" - we should just have the desktop behave as original (responsive design), since in most cases, they should behave the same since people aren't going to have a preview screen that's bigger the desktop width.
- Once you select a mode, the popover should disappear and then the popover trigger button (device mode button) should become highlighted if it's not in desktop mode, similar to what we do when you click on the Token Bar icon (this way people know they're in a non-default mode).

image

- We should use the shadcn tooltip and not the browser-native tooltip for the labels (e.g. Tablet, Mobile)

finally, the new e2e tests looks like they're not passing. could you take a look?

thank you.

wwwillchen (approved):
looks great, thank you!

PR #1620, opened 2025-10-23
feat: allow to update llm urls

### LINKED ISSUE #816: Allow to change the ollama and Lm Studio url
I'm currently limited to using Ollama's and Lm Studio local server, Right now, there's no way to change the endpoint or model URL, which makes it frustrating when I want to work with other locally hosted models like those in LM Studio.

I'd like to have an option in the settings that allows me to replace the default Ollama URL with a custom one, such as the LM Studio server URL (http://192.168.22.3:1234/v1/...). Ideally, this would be a configurable endpoint for local models.

Manually modifying hardcoded URLs in the source code.

Olyno:
# Description

This pull request introduces an interface to allow the users to update their LLM URLs inside Dyad.

Closes #816

image

    

---
## Summary by cubic
Add settings to edit local LLL endpoints (Ollama and LM Studio) and route all local model traffic through these values. Users can now point Dyad to remote or custom hosts while keeping sensible defaults.

- **New Features**
  - Provider page UI to view/update/reset Ollama and LM Studio endpoints.
  - Persist endpoints in UserSettings with localhost defaults.
  - LM Studio URL normalization: trim, add protocol, default port 1234, strip trailing /v1 and trailing slashes, IPv6-safe.
  - Ollama host parsing: trim, add protocol, default port 11434, IPv6-safe; use OLLAMA_HOST env if set, else settings.
  - Handlers and clients read settings for API base URLs (Ollama tags, LM Studio models, OpenAI-compatible provider); provider setup reflects local endpoints.
  - Added defaults as constants; tests for URL parsing/normalization and endpoint persistence; clearer connection errors.

<sup>Written for commit 7460516ea0b5ff03ba3557a66a6944a44ea43e0c. Summary will update on new commits.</sup>

---

> [!NOTE]
> Introduces configurable local model endpoints with persistence and consistent usage across the app.
> 
> - New `LocalModelEndpointSettings` UI on provider pages to view/update/reset `ollama` and `lmstudio` endpoints
> - Extend `UserSettings` with `ollamaEndpoint` and `lmStudioEndpoint` and default constants in `constants/localModels`
> - IPC: `local_model_ollama_handler` reads from env or settings with improved `parseOllamaHost`; `local_model_lmstudio_handler` uses `getLMStudioBaseUrl`; clearer connection errors
> - Utils: add `lm_studio_utils` for base URL normalization (protocol/port/trailing /v1 handling, IPv6-safe); `get_model_client` uses settings-derived base URLs for Ollama and LM Studio
> - UI/State: `ProviderSettingsPage` renders endpoint settings for local providers; `ProviderSettings` shows all providers; `useLanguageModelProviders` marks local providers as configured when endpoints are set
> - Tests: e2e for settings persistence; unit tests for LM Studio URL normalization, Ollama host parsing, and updated settings defaults
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7460516ea0b5ff03ba3557a66a6944a44ea43e0c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


wwwillchen (changes requested):
Thanks for the pull request.

A couple of high-level thoughts:

- Given that dyad is currently using the `OLLAMA_HOST` env var for ollama, is there a need to configure the ollama URL inside dyad? I'm not an expert in ollama, but it seems not super necessary to configure the ollama URL assuming we're picking up the env var correctly. For LM Studio, I could see this being useful as I'm not aware of any similar env var for it.
- UX-wise: I think we should configure this in the Model Providers section and not in the AI section. Similar to how each of the cloud providers have their own dedicated page, we should display the local model providers in this grid and then allow users to navigate to the provider-specific settings page and configure the URL:

image

elsung:
hm not sure if there are other folks like me who use ollama + tailscale, but for me theres a definite need so that i can use dyad on my laptop when im out and about to connect to my remote ollama instance

> Thanks for the pull request.
> 
> A couple of high-level thoughts:
> 
> * Given that dyad is currently using the `OLLAMA_HOST` env var for ollama, is there a need to configure the ollama URL inside dyad? I'm not an expert in ollama, but it seems not super necessary to configure the ollama URL assuming we're picking up the env var correctly. For LM Studio, I could see this being useful as I'm not aware of any similar env var for it.
> * UX-wise: I think we should configure this in the Model Providers section and not in the AI section. Similar to how each of the cloud providers have their own dedicated page, we should display the local model providers in this grid and then allow users to navigate to the provider-specific settings page and configure the URL:
> 
> image

Olyno:
> Given that dyad is currently using the OLLAMA_HOST env var for ollama, is there a need to configure the ollama URL inside dyad? I'm not an expert in ollama, but it seems not super necessary to configure the ollama URL assuming we're picking up the env var correctly. For LM Studio, I could see this being useful as I'm not aware of any similar env var for it.

As @elsung mentioned, the primary use-case here is to allow people to have remote instances, and not having to reload each time their Dyad instance whenever they need to change their Ollama instance url. I agree it shouldn't happen a lot, but this is nice to have, to be able to change the url directly from the dashboard.

> UX-wise: I think we should configure this in the Model Providers section and not in the AI section. Similar to how each of the cloud providers have their own dedicated page, we should display the local model providers in this grid and then allow users to navigate to the provider-specific settings page and configure the URL:

Got it, working on it!

Olyno:
Sorry for the delay! Would something like that work? @wwwillchen 

image

wwwillchen:
@Olyno no worries. I think the input should be inside the provider details page for Ollama and LM Studio respectively similar to how all the other providers work

Olyno:
@wwwillchen  Sounds good. Is something like that looking good, or would you prefer something else?

image

image

image

wwwillchen:
@Olyno looks good to me, thanks

Olyno:
Changes applied @wwwillchen 

:tada: 🥳 Happy new year 2026 🥳 :tada:

Let me know if there is anything which would require changes

wwwillchen:
@bugbot run

e2e-tests/local_provider_settings.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+testWithPo("Local provider endpoint settings persist", async ({ po }) => {
+  await po.setUp();
+  await po.goToSettingsTab();
+
+  await expect(po.page.getByText("Ollama", { exact: true })).toBeVisible();
+
+  await po.page.getByText("Ollama", { exact: true }).click();
+  await po.page.waitForSelector('h1:has-text("Configure Ollama")', {
+    state: "visible",
+    timeout: 5000,
```
wwwillchen, line 13: use Timeout.MEDIUM instead of hardcoding 5000 - same below
https://github.com/dyad-sh/dyad/blob/319d2b77b82c44f21d1eb2ce58ae58fa07acacaf/e2e-tests/helpers/test_helper.ts#L20

src/lib/schemas.ts
```diff
@@ -251,6 +251,8 @@ export type AgentToolConsent = z.infer<typeof AgentToolConsentSchema>;
 export const UserSettingsSchema = z.object({
   selectedModel: LargeLanguageModelSchema,
   providerSettings: z.record(z.string(), ProviderSettingSchema),
+  ollamaEndpoint: z.string(),
+  lmStudioEndpoint: z.string(),
```
wwwillchen, line 255: let's make these optional properties

src/main/settings.ts
```diff
@@ -23,6 +27,8 @@ const DEFAULT_SETTINGS: UserSettings = {
     provider: "auto",
   },
   providerSettings: {},
+  ollamaEndpoint: DEFAULT_OLLAMA_ENDPOINT,
```
wwwillchen, line 30: let's not set these here. instead, just use these default values when the field is undefined. otherwise, you'll need to rebaseline a lot of our e2e tests (see CI github actions)

wwwillchen (changes requested):
thanks for the changes! mostly looks good, a few minor changes and then i think we can merge

PR #1621, opened 2025-10-23
fix: update popup should not reset prompt

### LINKED ISSUE #1301: [bug] <Lost Entire Prompt Due to Update Pop-Up>
## Bug Description (required)
Not exactly a bug, and unlikely to happen often, but the update pop-up can appear while a large prompt is ready to be sent. If Enter is pressed at the same moment, the prompt is lost. This just happened to me, and it could be avoided by making Enter confirmation optional on the pop-up. No need to change how it appears.

## Steps to Reproduce (required)

1. Be unlucky
2. Press Enter just as the pop-up shows

## Expected Behavior (required)
The prompt should not be lost if the pop-up appears at an unlucky moment. 
Pressing Enter once should not automatically confirm the update.

## Actual Behavior (required)
The prompt is lost when Enter coincides with the pop-up appearing.

## System Information
- Dyad Version: 0.21.0
- Platform: win32
- Architecture: x64
- Node Version: v22.14.0
- PNPM Version: 10.11.0
- Node Path: C:\Program Files\nodejs\node.exe
- Telemetry ID: 85b01e11-945e-4c1e-ba5f-611e4ca823a4
- Model: google:gemini-2.5-pro | customId: undefined

## Logs
```
╭ Warning ─────────────────────────────────────────────────────────────────────╮
│                                                                              │
│   Ignored build scripts: @swc/core, esbuild.                                 │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed     │
│   to run scripts.                                                            │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯



> vite_react_shadcn_ts@0.0.0 dev C:\Users\gabri\dyad-apps\whispering-meerkat-sniff
> vite "--port" "32100"


  VITE v6.3.4  ready in 374 ms

  ➜  Local:   http://localhost:32100/
  ➜  Network: http://192.168.2.18:32100/


```

Olyno:
# Description

This pull request should avoid users to lose their prompts.

Closes #1301
    

---
## Summary by cubic
Prevents the update dialog from resetting in-progress prompts. After downloading an update, the app asks to restart instead of auto-restarting.

- **Bug Fixes**
  - Use update-electron-app onNotifyUser to show a Restart/Later dialog.
  - Restart only when confirmed (autoUpdater.quitAndInstall); otherwise keep the session.
  - Cleaned up imports and added MessageBoxOptions typing.

<sup>Written for commit a35d04a. Summary will update automatically on new commits.</sup>


wwwillchen:
@Olyno i don't understand how this fixes the issue. it seems functionally equivalent to the built-in dialog that electron updater has: https://github.com/electron/update-electron-app/blob/379618e570337c568cbadb89709f88a41134e7ad/src/index.ts#L256-L284 - could you explain how this is different from the current behavior?

Olyno:
I agree, it can be quite confusing at first. Basically, instead of making the "accept update" as default, it becomes the "close" action, which avoids running the update, and losing the prompt. 

In short, instead of accepting the update when pressing Enter, we simply close the update popup. @wwwillchen

wwwillchen:
@Olyno sorry for the slow response. i don't think this is the right approach because then users may inadvertently skip updates, and some of them are quite important for fixing bugs, etc.
instead, i think we should probably have a more discrete upgrade UI (similar to Cursor and other products) where a banner tells users that an upgrade is available and they can choose when to upgrade.

Issue #1591, opened 2025-10-20
Allow referencing files (like prompts/apps) to be used for context


wwwillchen:

azizmejri1:
Hi @wwwillchen ! I’m interested in working on this issue, just checking whether you or someone else is already working on it before I start.

wwwillchen:
Nope, feel free to work on this!

PR #1648, opened 2025-10-27
Feat : allow referencing files

azizmejri1:
I implemented file referencing feature mentioned in issue #1591 

---
## Summary by cubic
Adds support for referencing app files in chat using @file:<path>. The input now autocompletes files, and the backend validates and surfaces referenced files in the chat context as read-only.

- **New Features**
  - Chat input autocompletes @ mentions for apps, prompts, and files.
  - Recognizes @file:<path> and converts it to a structured mention on submit.
  - Backend parses @file mentions, checks file existence, and adds a “Referenced Files” section to the system message.
  - New get-app-files IPC handler and useAppFiles hook to load file paths for the selected app (appFilesAtom added).
  - e2e test for mentioning a file and capturing server dump.


src/hooks/useAppFiles.ts
wwwillchen, line 1: You can just use the existing `useLoadApp` hook - which fetches the app, which has the list of filenames under the `files` property (so you don't need to add a new hook, IPC, etc.)
https://github.com/dyad-sh/dyad/blob/8a3bc53832f082cfc6ac18d7b508c50e37ae901b/src/hooks/useLoadApp.ts#L8

src/ipc/handlers/chat_stream_handlers.ts
```diff
@@ -340,6 +340,37 @@ export function registerChatStreamHandlers() {
...
+      // --- File mention support: parse @file:<path> ---
+      let referencedFilesSection = "";
+      try {
+        const fileMentionRegex = /@file:([^\s]+)/g;
+        const mentionedFiles = [];
+        let match;
+        while ((match = fileMentionRegex.exec(userPrompt)) !== null) {
+          mentionedFiles.push(match[1]);
+        }
+        console.log(mentionedFiles);
```
wwwillchen: remove

src/ipc/handlers/chat_stream_handlers.ts
```diff
@@ -340,6 +340,37 @@ export function registerChatStreamHandlers() {
...
+        if (mentionedFiles.length > 0) {
+          const appRoot = getDyadAppPath(chat.app.path);
+          for (const relPath of mentionedFiles) {
+            // Prevent directory traversal
```
wwwillchen: use `safeJoin` from src/ipc/utils/path_utils.ts

src/ipc/handlers/chat_stream_handlers.ts
```diff
@@ -573,6 +604,10 @@ ${componentSnippet}
 
           systemPrompt += `\n\n# Referenced Apps\nThe user has mentioned the following apps in their prompt: ${mentionedAppsList}. Their codebases have been included in the context for your reference. When referring to these apps, you can understand their structure and code to provide better assistance, however you should NOT edit the files in these referenced apps. The referenced apps are NOT part of the current app and are READ-ONLY.`;
         }
+        // Add referenced files section if any
```
wwwillchen: i know you're following the pattern in referenced apps, but i think we can just remove this (and keep the system prompt as-is) - I think LLMs, will automatically realize what's going on and focus on the files mentioned in the prompt. the reason we had a system prompt for the referenced apps, is that I was more concerned the LLM would get confused and try to edit referenced apps (which is not allowed)

wwwillchen (changes requested):
overall, nice job! i think we can simplify the code by using some of the existing utils/modules, but UX-wise this looks right

src/ipc/handlers/chat_stream_handlers.ts
azizmejri1: Hi @wwwillchen , if i'm not gonna add referenced files to the system prompt should i just remove the whole part from L343 to L373?

wwwillchen: yeah

src/components/chat/LexicalChatInput.tsx
```diff
@@ -271,16 +279,24 @@ export function LexicalChatInput({
...
+      type: "app",
+    }));
 
     const promptItems = (prompts || []).map((p) => ({
       value: p.title,
       type: "prompt",
       id: p.id,
     }));
 
+    const fileItems = (appFiles || []).map((item) => ({
```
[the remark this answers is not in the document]
wwwillchen: please do this suggestion.

wwwillchen (changes requested):

wwwillchen (approved):
nice, @azizmejri1! thank you

PR #1637, opened 2025-10-26
Support reusable modules for Supabase edge functions (Closed due to rebase edits removing all past commits so I could start on a clean slate free of esbuild)

coderdylanbates:
Screenshot 2025-10-26 180622

Closes #1630 

---
## Summary by cubic
Enable reusable modules for Supabase Edge Functions by bundling each function with code from supabase/functions/_shared and auto-redeploying affected functions when shared helpers change.

- New Features
  - Bundle functions with esbuild (ESM, externalize non-relative/URL imports); require index.ts per function.
  - Auto-deploy on write/rename/delete: deploy only the touched function; if a shared file changes, redeploy all functions.
  - New Supabase section in Settings shows detected function folders and shared module files.
  - Updated prompt and added README to guide using supabase/functions/_shared.
  - Added tests for bundling and path utilities.

- Bug Fixes
  - Prevent Electron main build errors by marking esbuild as external in vite.main.config.mts.

PR #1665, opened 2025-10-29
Support reusable modules for Supabase edge functions (Draft) (Last PR was closed automatically due to commits being removed from rebase)

### LINKED ISSUE #1630: Support reusable modules for Supabase edge functions
https://chatgpt.com/share/68fc1389-c628-8007-9c23-f271970f9cf8

Use `_shared` to avoid accidental deployments with supabase CLI and respect convnetions.

coderdylanbates:
Closes #1630 

This PR is a duplicate of #1637 because Github closed that when I rebased to clean the slate from esbuild.


coderdylanbates:
Turns out Supabase CLI package is different from the supabase-js package, so unless the users install an extra supabase package they cannot use the native supabase method so I've just changed it to use zips for bundling.

src/supabase_admin/supabase_management_client.ts
wwwillchen, line 1: Thanks for the PR - it's taking me a while to review this file and I was wondering if you can explain to me why you took this approach. Does Supabase Edge function require a single file (zip) to upload? 

When I first saw https://supabase.com/docs/reference/api/v1-deploy-a-function, I thought it would take multiple files since `file` is an `Array<string>` but now I'm wondering if I was wrong and it takes a single file. 

I tried reading through what Supabase CLI does https://github.com/supabase/cli/blob/891a5df913832f4a99ee24ca1ac515d068b8f80b/internal/functions/deploy/deploy.go#L4  and it seems like they ultimately shell out to some bundler like what you did with esbuild.

I need to think a bit more about what the right approach is because I think bundling `esbuild` into Dyad would be too heavy handed, but potentially we could use the `esbuild` inside the user's app directory node_modules (esp. if they are using Vite, which is the case for most Dyad users using supabase).

What do you think? It'd be good to hear your thoughts on the pros and cons of the different options as this is quite a tricky problem (and harder than I expected!)

coderdylanbates: I've changed it to just directly send files without using esbuild or zip, I'm sorry for the confusion on those commits you saw previously. Those commits before were a rough draft of the updated functionality which is why I marked the branch as draft, however it is more stable and ready to review now.

src/supabase_admin/supabase_management_client.ts
```diff
@@ -260,13 +412,279 @@ export async function deploySupabaseFunctions({
...
+  }
+  return table;
+})();
+
+function crc32(buffer: Buffer): number {
+  let crc = 0 ^ -1;
+  for (let i = 0; i < buffer.length; i++) {
+    crc = (crc >>> 8) ^ CRC32_TABLE[(crc ^ buffer[i]) & 0xff];
+  }
+  return (crc ^ -1) >>> 0;
```
wwwillchen: i don't think this is needed anymore

src/supabase_admin/supabase_management_client.ts
wwwillchen, line 1: i think the zip-related logic in this file can now be removed (doesn't look like it's needed)

wwwillchen (changes requested):
sorry for the delayed response. thanks for making these changes! I pulled it down and it works well overall :)

A couple of high-level feedback:
1. If a file in the shared directory gets modified by dyad, currently in this PR, the supabase edge functions which depend on the shared module don't re-deploy. to keep things simple, i think we could just re-deploy all edge functions whenever something in the shared directory is modified.
2. by convention, i think usually the shared directory for supabase edge functions is located at `supabase/functions/_shared` ([example](https://github.com/supabase/supabase/tree/master/examples/edge-functions/supabase/functions/_shared)), right now it's at `supabase/shared` which works but it's a bit unconventional and i think generally following convention is good because that's what AI wants to do (since it sees it in the training data)

I think once these two things are addressed and the code is cleaned up (see inline comments below), we can merge this. thanks!

wwwillchen:
@SlayTheDragons just checking if this is ready for review?

coderdylanbates:
Sorry for delays I've been busy with exams but I should have some more time available to work on it now.

wwwillchen:
@SlayTheDragons no worries, just mark the PR as open once it's ready, thx

wwwillchen:
thanks @SlayTheDragons for this PR, I've opened a new PR which is based on this PR https://github.com/dyad-sh/dyad/pull/1964 (I found it a little hard to modify this PR directly due to rebasing/version conflicts).

thank you for exploring different options, it was very helpful!

Issue #1666, opened 2025-10-29
Don't show Upgrade to Pro with `TypeError: terminated`


wwwillchen:
https://www.reddit.com/r/dyadbuilders/comments/1oj5ump/pro/

PR #1667, opened 2025-10-29
fix : hiding upgrade banner on unrelated errors

azizmejri1:
This PR solves the issue #1666 
To solve this i removed the upgrade link from the general case since errors related to pro mode are already handled in the specific if-blocks

---
## Summary by cubic
Hide the Dyad Pro upgrade banner on termination errors to reduce confusion. The generic upgrade link in ChatErrorBox is now suppressed when the error includes "TypeError: terminated" and only shows when Pro is disabled otherwise.

<sup>Written for commit 1edf665. Summary will update automatically on new commits.</sup>

### DIFF
--- a/src/components/chat/ChatErrorBox.tsx
+++ b/src/components/chat/ChatErrorBox.tsx
@@ -99,6 +99,14 @@ export function ChatErrorBox({
     <ChatErrorContainer onDismiss={onDismiss}>
       {error}
       <div className="mt-2 space-y-2 space-x-2">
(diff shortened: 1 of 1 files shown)


wwwillchen:
thanks for the pr @azizmejri1  - we should still allow the general case to show the pro upgrade button because there's many errors that aren't covered but we should show it (e.g. users hitting rate limit w/ openrouter API). instead, let's just check if the error message contains `TypeError: terminated` and if it does, we don't show the pro button.

wwwillchen:
thanks!

PR #1728, opened 2025-11-07
feat: multi-component-selector

azizmejri1:
## Summary by cubic
Adds multi-component selection in the preview and sends all selected components to chat for targeted edits. Updates overlays, UI, and IPC to support arrays, smarter context focusing, and cross-platform path normalization.

- **New Features**
  - Select multiple components in the iframe; selection mode stays active until you deactivate it.
  - Show a scrollable list of selections with remove buttons and a Clear all; remove from the list or click an overlay in the preview to deselect. Sending clears all overlays.
  - Separate hover vs selected overlays with labels on hover; overlays persist after deactivation and re-position on layout changes/resizes.
  - Chat input and streaming now send selectedComponents; server builds per-component snippets and focuses their files in smart context.

- **Migration**
  - Replace selectedComponentPreviewAtom with selectedComponentsPreviewAtom (ComponentSelection[]).
  - ChatStreamParams now uses selectedComponents; migrate any single-selection usages.
  - previewIframeRefAtom added for clearing overlays from the parent.

<sup>Written for commit da0d64cc9e9f83fbf4b975278f6c869f0d3a8c7d. Summary will update automatically on new commits.</sup>


wwwillchen:
nice work @azizmejri1! I think if there's a way to de-select a component from the preview itself that would be helpful. for example, if i click on a selected component again, it should de-select it. this feels like a pretty natural user interaction pattern.

image


the other suggestion would be to have some way to de-select all the components. for example, if there's more than 1 component selected, could we have a button at the top like Clear all selected components in the chat input area:

image

azizmejri1:
Thanks for the feedback @wwwillchen!
I’m actually working on those two improvements. Since this PR is already pretty big, should I include them in a separate one?

wwwillchen:
@azizmejri1 cool - I'm OK with keeping it in this PR or a separate PR. in general, i'm OK with reviewing big PRs as long as they're related changes. thanks

azizmejri1:
@wwwillchen This feature is almost done. I’m running into a small issue with snapshots , since I’m on Windows, I can’t generate the Mac snapshots (There’s a path slash difference (\ vs /) causing some diffs)

wwwillchen:
@azizmejri1 use normalizePath so that it's consistent across windows and macs https://github.com/dyad-sh/dyad/blob/cbf247a4efb168ea5b906cc60bd7feefbe76b56d/shared/normalizePath.ts#L8

e2e-tests/select_component.spec.ts
```diff
@@ -27,6 +27,34 @@ testSkipIfWindows("select component", async ({ po }) => {
...
+    .click();
+
+  await po.snapshotPreview();
+  await po.snapshotSelectedComponentDisplay();
+
+  await po.sendPrompt("[dump] make both smaller");
+  await po.snapshotPreview();
+  await expect(po.getSelectedComponentDisplay()).not.toBeVisible();
+
+  await po.snapshotServerDump("all-messages");
```
wwwillchen: I think we just need to dump `"last-message"` since the other messages aren't really relevant

src/ipc/ipc_client.ts
```diff
@@ -388,6 +388,7 @@ export class IpcClient {
     prompt: string,
     options: {
       selectedComponent: ComponentSelection | null;
```
wwwillchen: can we remove `selectedComponent` param?

worker/dyad-component-selector-client.js
wwwillchen, line 1: can we do something where when the mouse leaves the window, to not show the overlay? otherwise, it's a bit of a weird UX where the outer-most component retains the overlay even though the user might be typing in the prompt/chat input, etc.

image

src/hooks/useStreamChat.ts
```diff
@@ -70,13 +70,15 @@ export function useStreamChat({
       redo,
       attachments,
       selectedComponent,
```
wwwillchen: same as elsewhere- let's remove `selectedComponent` and just use `selectedComponents`

wwwillchen (changes requested):
overall, very nice! left a few comments.

wwwillchen (approved):
@azizmejri1 great work! looks good. 

on a side note: if you can resolve comments once you've addressed them, that would be helpful and it makes it clearer and cleaner what's been done. thanks!

PR #1760, opened 2025-11-12
Replace native Git with Dugite to support users without Git installed

princeaden1:
I moved all isomorphic-git usage into a single git_utils.ts file and added Dugite as an alternative Git provider. The app now checks the user’s settings and uses dugite when user enabled native git for all isomorphic-git commands. This makes it easy to fully remove isomorphic-git in the future by updating only git_utils.ts.

---

> [!NOTE]
> Adds Dugite-based native Git (bundled binary) and refactors all Git calls to a unified git_utils API, replacing direct isomorphic-git usage across the app.
> 
> - **Git Platform Abstraction**:
>   - Introduces `dugite` and bundles Git via Electron Forge (`extraResource`) with `LOCAL_GIT_DIRECTORY` setup in `src/main.ts`.
>   - Adds `src/ipc/git_types.ts` and a comprehensive `src/ipc/utils/git_utils.ts` wrapper supporting both Dugite (native) and `isomorphic-git` (fallback): `commit`, `add`/`addAll`, `remove`, `init`, `clone`, `push`, `setRemoteUrl`, `currentBranch`, `listBranches`, `renameBranch`, `log`, `isIgnored`, `getCurrentCommitHash`, `getGitUncommittedFiles`, `getFileAtCommit`, `checkout`, `stageToRevert`.
> - **Refactors (switch to git_utils)**:
>   - Replaces direct `isomorphic-git` imports in handlers and processors: `app_handlers`, `chat_handlers`, `createFromTemplate`, `github_handlers`, `import_handlers`, `portal_handlers`, `version_handlers`, `response_processor`, `neon_timestamp_utils`, `utils/codebase`.
>   - Updates tests to mock `git_utils` (`src/__tests__/chat_stream_handlers.test.ts`).
> - **Behavioral/Feature Updates**:
>   - `createFromTemplate` uses `fetch` for GitHub API and `gitClone` for cloning with cache validation.
>   - GitHub integration uses `gitSetRemoteUrl`/`gitPush`/`gitClone`, handling public vs token URLs and directory creation when native Git is disabled.
>   - Versioning, imports, app file edits, migrations now stage/commit via `git_utils`.
> - **UI/Copy**:
>   - Updates Settings description for “Enable Native Git”.
> - **Config/Version**:
>   - Bumps version to `0.29.0-beta.1`; adds `dugite` dependency.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ba098f7f25d85fc6330a41dc718fbfd43fff2d6c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


wwwillchen:
Thanks @princeaden1 - a couple of high-level questions/feedback:

1. Do you know how much larger adding Dugite will make the Dyad binary/zip file?
2. I think we should wrap all the dugite vs. isomorphic-git calls into `git_utils.ts` this makes it cleaner if other places want to call these standard git commands.

princeaden1:
@wwwillchen thanks for the feedback!
1. Based on my findings, the Dugite Git binary size varies by platform. The macOS bundle is the largest at roughly 40–55 MB, while Windows and Linux binaries range between 25–50 MB. 
To avoid shipping all platform binaries inside a single Dyad build, we'll only include the binary for the target OS at build time. This keeps the final Dyad binary/zip size minimal and avoids bundling unnecessary Git binaries for other platforms.

2. Yes, I'll wrap all Dugite vs isomorphic-git interactions inside git_utils.ts so all git operations go through a single utility layer.

src/ipc/utils/git_utils.ts
```diff
@@ -1,63 +1,62 @@
...
-    if (error.stdout) {
-      errorMessage += `\nStdout: ${error.stdout}`;
-    }
-    if (error.stderr) {
-      errorMessage += `\nStderr: ${error.stderr}`;
-    }
-    throw new Error(errorMessage);
-  }
-}
+import type {
```
wwwillchen, line 11: instead of defining these new types in ipc_types, I'd define it in a new file e.g. git_types.ts

The purpose of ipc_types.ts is for types that are shared between the client-side and Node.js JS contexts, particularly for ipc_client & it's associated handlers, but I think these git types are only used o the Node.js types

src/ipc/utils/git_utils.ts
```diff
@@ -170,30 +160,89 @@ export async function gitStageToRevert({
...
+  } else {
+    await git.remove({
+      fs,
+      dir: path,
+      filepath,
+    });
+  }
+}
+
+export async function gitStatus({ path }: GitBaseParams): Promise<string[]> {
```
wwwillchen: let's give this a more specific name like getGitUncommittedFiles

src/ipc/utils/git_utils.ts
```diff
@@ -221,3 +270,285 @@ export async function getFileAtCommit({
...
+    });
+  }
+}
+
+export async function gitClone({
+  path,
+  url,
+  accessToken,
+  singleBranch = true,
+  depth = 1,
```
wwwillchen: let's use `undefined` as the default value instead of  `1` to match git clone's behavior (seems more intuitive to me)

wwwillchen (changes requested):
Thanks @princeaden1! Good work.

Couple of comments:
1. I noticed the e2e tests for native git are failing - could you look into why they're failing? Also, I think it would be good to add one or two e2e test cases for native git + github in e2e-tests/github.spec.ts
2. The text "Native Git offers faster performance but requires installing Git." - should be updated as we longer require installing Git (since dugite is pre-packaged) https://github.com/dyad-sh/dyad/blob/8dee2552bb3298404ddc9a940e1f5b62b3d5a4ca/src/pages/settings.tsx#L166-L176

princeaden1:
@wwwillchen I've updated the PR. I also noticed that the GitHub workflow is failing, looks like an issue with the git/pnpm cache step.

wwwillchen:
@princeaden1 i re-ran the failed GitHub jobs, not sure what was happening - it looks like it was a transient error, but it looks like there's still several e2e tests failing with messages like:

>   pw:browser [pid=28016][err] 01:50:55.923 (import-handlers) › Error in : args: [{"path":"/Users/runner/work/dyad/dyad/e2e-tests/fixtures/import-app/version-integrity","appName":"version-integrity-app"}] Error: ENOENT: Git failed to execute. This typically means that the path provided doesn't exist or that the Git executable could not be found which could indicate a problem with the packaging of dugite. Verify that resolveGitBinary returns a valid path to the git binary. +24ms

princeaden1:
> @princeaden1 i re-ran the failed GitHub jobs, not sure what was happening - it looks like it was a transient error, but it looks like there's still several e2e tests failing with messages like:
> 
> > pw:browser [pid=28016][err] 01:50:55.923 (import-handlers) › Error in : args: [{"path":"/Users/runner/work/dyad/dyad/e2e-tests/fixtures/import-app/version-integrity","appName":"version-integrity-app"}] Error: ENOENT: Git failed to execute. This typically means that the path provided doesn't exist or that the Git executable could not be found which could indicate a problem with the packaging of dugite. Verify that resolveGitBinary returns a valid path to the git binary. +24ms




@wwwillchen Thanks,  for re-running those jobs.

I've checked the failing tests in my local dev environment and they’re all passing. I think this is related to the path for node_modules/dugite on the GitHub.

I'll dig into it and come back to you with full details, including whether it will require any code changes.

princeaden1:
I found what is most likely causing the error, will fixed this today.

> > @princeaden1 i re-ran the failed GitHub jobs, not sure what was happening - it looks like it was a transient error, but it looks like there's still several e2e tests failing with messages like:
> > > pw:browser [pid=28016][err] 01:50:55.923 (import-handlers) › Error in : args: [{"path":"/Users/runner/work/dyad/dyad/e2e-tests/fixtures/import-app/version-integrity","appName":"version-integrity-app"}] Error: ENOENT: Git failed to execute. This typically means that the path provided doesn't exist or that the Git executable could not be found which could indicate a problem with the packaging of dugite. Verify that resolveGitBinary returns a valid path to the git binary. +24ms
> 
> @wwwillchen Thanks, for re-running those jobs.
> 
> I've checked the failing tests in my local dev environment and they’re all passing. I think this is related to the path for node_modules/dugite on the GitHub.
> 
> I'll dig into it and come back to you with full details, including whether it will require any code changes.

wwwillchen:
thanks @princeaden1 - there's a TS error (https://github.com/dyad-sh/dyad/actions/runs/19976334752/job/57293296263?pr=1760):

> Error: src/ipc/utils/git_utils.ts(51,30): error TS2304: Cannot find name 'execAsync'.

src/main.ts
```diff
@@ -41,6 +41,27 @@ registerIpcHandlers();
 if (started) {
   app.quit();
 }
+// Set up Dugite git path for development
```
wwwillchen: let's remove this

src/main.ts
```diff
@@ -41,6 +41,27 @@ registerIpcHandlers();
...
+// if (!app.isPackaged && !process.env.CI) {
+// <-- ADDED CHECK
+// Set up Dugite git path for development and e2e tests
+if (!app.isPackaged || IS_TEST_BUILD) {
+  // For tests, use process.cwd() which should be the project root
```
wwwillchen: how come test shouldn't use app.getAppPath()?

I'm wondering how this is going to work in the actual app build (which is similar to the e2e test)

src/main.ts
```diff
@@ -41,6 +41,27 @@ registerIpcHandlers();
...
+  // For dev, use app.getAppPath() which is also the project root
+  const projectRoot = IS_TEST_BUILD ? process.cwd() : app.getAppPath();
+  const gitPath = path.join(projectRoot, "node_modules/dugite/git");
+
+  const fs = require("fs");
```
wwwillchen: this fallback behavior seems kind of weird. could you explain why it's needed?

wwwillchen (changes requested):
have a couple questions about main.ts

princeaden1:
> thanks @princeaden1 - there's a TS error (https://github.com/dyad-sh/dyad/actions/runs/19976334752/job/57293296263?pr=1760):
> 
> > Error: src/ipc/utils/git_utils.ts(51,30): error TS2304: Cannot find name 'execAsync'.

Thanks @wwwillchen for the feedback,
1. Regarding the execAsync error. execAsync has been completely removed in this PR,  the error was caused because I merge the new update for the deep context which has added isGitStatusClean. I have updated the code now and the error has now been fixed. Thanks.

2. Regarding the main.ts Git path logic, I've updated the logic to make the environment handling clearer.
For development, app.getAppPath() resolves to a different location from the test environment, so I needed an different path check (process.cwd()) to resolve the correct node_modules/dugite/git path on test environment. app.getAppPath() works fine for the development. For production, app.getAppPath() points inside Resources/app.asar, which I think will not contain node_modules. I have used Electron Forge bundles Git into Resources/git via extraResource, so using getAppPath() for prod could break production. The updated code is seems to be more clearer. 


```ts|function resolveLocalGitDirectory() {
  if (IS_TEST_BUILD) {
    // Tests run from repo root
    return path.join(process.cwd(), "node_modules/dugite/git");
  }

  if (!app.isPackaged) {
    // Dev: app.getAppPath() is the project root
    return path.join(app.getAppPath(), "node_modules/dugite/git");
  }

  // Packaged app: git is bundled via extraResource
  return path.join(process.resourcesPath, "git");
}

const gitDir = resolveLocalGitDirectory();
if (fs.existsSync(gitDir)) {
  process.env.LOCAL_GIT_DIRECTORY = gitDir;
}```

File: src/main.ts [hunk 7]
```diff
@@ -42,6 +43,22 @@ if (started) {
   app.quit();
 }
 
+// Decide the git directory depending on environment
+function resolveLocalGitDirectory() {
+  if (!app.isPackaged) {
+    // Dev: app.getAppPath() is the project root
+    return path.join(app.getAppPath(), "node_modules/dugite/git");
+  }
+
```
wwwillchen, line 52: @princeaden1 i patched it down and tried using process.resourcesPath for the e2e test and it worked. i've simplified the code so it uses the same path for packaged app regardless of e2e testing or regular app distribution.

princeaden1: @wwwillchen  sounds great. I;ll go ahead and fix the errors prompt by cursors now. thanks

wwwillchen (approved):

Issue #1636, opened 2025-10-26
Edit button


coderdylanbates:
⚡ I'm always annoyed when I ask the AI for something and it doesn't quite work because I didn't specify something properly and then I have to contradict what I said and the AI gets confused because I'm using a cheap model,

✅  Therefore, I'd love if we could add an **edit button** on the messages like ChatGPT has that rolls back the conversation to the point of the edited message, it would still keep the codebase how it was before however.

If you like I'll make a PR for this if you assign it to me, I just opened this issue so I can clarify if this is wanted feature for dyad or not. 😃

coderdylanbates:
Opps this already existed

Issue #1231, opened 2025-09-09
✏️ Let Us Edit Prompts & 📋 Copy Responses – Missing Basics!


heyvoon:
**Is your feature request related to a problem? Please describe.**  
I'm always frustrated when I realize I made a typo or want to refine my prompt after already submitting it to the AI. Currently, I have to start a new conversation or retype the entire prompt, which breaks the flow. Additionally, I often want to copy the AI’s response to use elsewhere (e.g., notes, messages, documents), but there’s no easy way to do so—long-pressing doesn’t always work reliably, and formatting can get lost.

**Describe the solution you'd like**  
1. **Editable Prompts**: Allow users to edit their submitted prompts after sending. This could be a small pencil/edit icon next to the user’s message that lets them revise the text. The AI would then update its response based on the edited prompt, ideally with a clear indicator that the prompt was changed.  
2. **Copy Button**: Add a “Copy” button (📋) next to each of Dyad’s responses. Tapping it should copy the full text of the response to the clipboard with formatting preserved (e.g., code blocks, lists, etc.), and show a brief confirmation (e.g., “Copied!”).

**Describe alternatives you've considered**  
- For editing: I’ve tried manually re-sending corrected prompts, but this clutters the chat and loses context.  
- For copying: Manually selecting and copying the response doesn’t preserve formatting—when pasted, the text becomes a disorganized block without proper line breaks or structure. 
- As an alternative, a “Regenerate” option based on an edited prompt could work, but direct prompt editing feels more intuitive and seamless.

**Additional context**  
These features are common in other AI chat apps (like ChatGPT or Claude), and their absence in Dyad makes the experience feel less polished. Being able to edit prompts would encourage experimentation and reduce errors. A copy button would greatly improve usability, especially for developers or writers who frequently reuse AI-generated content. A simple UI addition (edit and copy icons on hover or tap) would go a long way in improving user experience.

oliviakate5244-bit:
Thanks for reporting this! We will review where it fits in our priorities.

princeaden1:
@wwwillchen I have a working system for this, it requires adding a new column to the chat database to track the edits, I don't know if it is something you would like to have. I can push a PR for this if you dont mind.

wwwillchen:
@princeaden1 feel free to open a PR - thx

PR #1818, opened 2025-11-18
feat: Edit + Copy Chat Prompt with Versioning (#1231) (#1636)

princeaden1:
Edit Prompt and Copy Prompt functionality with full version tracking. When a user edits a message, a new version is created while the previous version is hidden. Users can navigate between versions using the version controls.

**Key Changes**
- Added edit prompt flow with versioned message creation
- Implemented copy prompt action
- Ensured edited messages generate new version entries without modifying the original
- Added UI logic for showing/hiding versions

**Next Steps**
Finalize E2E tests (to be completed in a follow-up PR)

Closes
(#1231) (#1636)

### DIFF
--- a/drizzle/0017_faithful_slapstick.sql
+++ b/drizzle/0017_faithful_slapstick.sql
@@ -1,0 +1,3 @@
+ALTER TABLE `messages` ADD `parent_message_id` integer;--> statement-breakpoint
+ALTER TABLE `messages` ADD `version_number` integer DEFAULT 1 NOT NULL;--> statement-breakpoint
+ALTER TABLE `messages` ADD `branch_id` text;
--- a/drizzle/meta/0017_snapshot.json
+++ b/drizzle/meta/0017_snapshot.json
@@ -1,0 +1,782 @@
+{
+  "version": "6",
+  "dialect": "sqlite",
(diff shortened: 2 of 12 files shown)


wwwillchen (changes requested):
i think this is an interesting idea, but there's a lot of weird edge cases right now:
- this PR breaks existing user chat messages - it doesn't render old user chat messages.
- if you go to an earlier message, and then send a new message. the new message is added under the latest message, which is a confusing UX
- if you edit a message, it doesn't undo the previous change.

if you look at what editors like Cursor do, I think their editing experience is more intuitive (and simpler).
- if you edit, it effectively does a restore version to the state before that user message.
- you can't cycle between the old and new versions.

i think chatgpt's UX makes sense b/c it's a pure chat environment but it gets very confusing once you deal with code.

wdyt?

princeaden1:
@wwwillchen You're totally right, thanks for calling out those cases. I'm reworking the code now to align with your suggestion. I initially thought we could lean more toward the ChatGPT-style flow, but as you mentioned, that UX fits pure chat much better than an agentic/code-editing context.

I'll update the PR this week to reflect this direction.

princeaden1:
@wwwillchen, I am closing this PR, I'll open a new PR for a much more simpler version of Edit Prompt Feature.

PR #1828, opened 2025-11-19
Visual editing

azizmejri1:
This is a prototype of the visual editing feature

---
## Summary by cubic
Prototype visual editing mode for the preview app. Toggle the mode, pick elements (single or multiple), and edit margin, padding, border, background, static text, and text styles with live updates, then save changes back to code.

- **New Features**
  - Pen tool button to enable/disable visual editing in the preview and toggle single/multi select; pro-only.
  - Inline toolbar anchored to the selected element for Margin (X/Y), Padding (X/Y), Border (width/radius/color), Background color, Edit Text (when static), and Text Style (font size/weight/color/font family).
  - Reads computed styles from the iframe and applies changes in real time; auto-appends px; overlay updates on scroll/resize.
  - Save/Discard dialog batches edits and writes Tailwind classes to source files via IPC; uses AST/recast to update className and text, replacing conflicting classes by prefix; supports multiple components.
  - New visual editor worker to get/apply styles and enable inline text editing via postMessage; selector client updated for coordinates streaming and highlight/deselect.
  - Proxy injects the visual editor client; new atoms track selected component, coordinates, and pending changes; component analysis flags dynamic styling and static text.
  - Uses runtimeId to correctly target and edit duplicate components.

- **Dependencies**
  - Added @babel/parser for AST-based text updates.
  - Added recast for safer code transformations.

<sup>Written for commit cdd50d33387a29103864f4743ae7570d64d61e93. Summary will update automatically on new commits.</sup>


wwwillchen (commented):
Thanks for sharing this draft!

a few thoughts after playing with this:
- I think it's kind of weird to have two buttons that are both selecting components. I think there should be one button (similar to what lovable does) and then you can either use it to chat with AI _or_ do visual editing.
- we should support editing text, changing font size, etc. i'm not sure about changing width and height directly though - if the element is directly sized, then it may make sense, but if the width and height is computed based on text content for example, then it shouldn't be configurable.
- i think a horizontal toolbar is more intuitive IMO

overall, nice work and looks like a promising start

azizmejri1:
@wwwillchen This feature is almost ready, it just needs some additional UI/UX improvements.

Right now, users can choose between selecting a single component or multiple components. However, I’ve noticed that when multi-select is disabled, the visual editing interface looks much cleaner.

To improve the experience, we could allow visual editing only when multi-select is turned off. For example, tools like Lovable only enable visual editing when a single component is selected. 

Any thoughts on this ?

wwwillchen:
thanks @azizmejri1! this is shaping up quite nicely.

a few thoughts:
- i think making the user select between single select and multiple select adds a bit too much friction and sometimes users aren't sure whether they want to select single or multiple yet. i think we should just have a single unified (multiselect) mode. - although it's a little confusing UX-wise when there's multiple elements selected + visual editor, I think users will still figure it out - we could always change this later based on feedback during beta testing.
- i think we should commit the visual editor changes - right now, we just write the file changes but don't commit it.
- b/c you're using babel to generate the file from the AST, it causes a lot of formatting changes - this isn't a dealbreaker but somewhat annoying for users who look at the code :) - could you look at using https://github.com/benjamn/recast which would preserve the existing formatting (for the parts of teh AST that don't change)?

wwwillchen:
@azizmejri1 one more thought about this: I think we should Visual Editing a pro-only feature because it's powerful but also not essential to building apps with dyad. 

we should:
1. have a runtime check that the user is a valid Dyad Pro user - see https://github.com/dyad-sh/dyad/blob/2ca14345b67bcfdb4a241a9ad6f91822d62480d8/src/components/chat/MessagesList.tsx#L38 - as long as  userBudget is a truthy object, then the user has put in a valid Dyad Pro key (it's a little confusing to use this as the pro check, but sometimes users put in invalid Dyad Pro keys and we don't want to check whether Dyad Pro is enabled because sometimes Pro users will want to disable using Dyad Pro and use their own API keys for model requests - since visual editing doesn't really consume any AI credits, Dyad Pro customers should always be able to use visual editing).
2. move some of the files (maybe just visual_editing_handlers.ts) to https://github.com/dyad-sh/dyad/tree/main/src/pro - the `main` folder is for code running in node.js, so you could put visual_editing_handlers.ts there.

azizmejri1:
@wwwillchen Alright, I’ll work on this.

e2e-tests/snapshots/visual_editing.spec.ts_visual-editing-single-component-margin.txt
```diff
@@ -0,0 +1,18 @@
+=== src/pages/Index.tsx ===
+// Update this page (the content is just a fallback if you fail to update the page)
+import { MadeWithDyad } from "@/components/made-with-dyad";
+
+const Index = () => {
+    return (
+        <div className="min-h-screen flex items-center justify-center bg-gray-100">
+            <div className="text-center">
+                <h1 className="text-4xl font-bold mb-4 mx-[20px] my-[10px]">Welcome to Your Blank App</h1>
```
wwwillchen: although this test snapshot is technically correct, i think it's a little problematic we generate code like this `mb-4 mx-[20px] my-[10px]` - i think we should clear out the mb-4 if the user edits the y margin

e2e-tests/visual_editing.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  // Find the heading element in the iframe
+  const heading = frame.getByRole("heading", {
+    name: "Welcome to Your Blank App",
+  });
+
+  await heading.dblclick();
+
+  // Wait a bit for contentEditable to be enabled
+  await po.page.waitForTimeout(300);
```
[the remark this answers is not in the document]
wwwillchen: @azizmejri1 let's do a more explicit wait like this:

```ts
await expect(async () => {
  const isEditable = await element.evaluate(el => el.isContentEditable);
  expect(isEditable).toBe(true);
}).toPass({ timeout: ... });
```

src/components/chat/ChatInput.tsx
```diff
@@ -306,6 +321,32 @@ export function ChatInput({ chatId }: { chatId?: number }) {
               />
             )}
 
+          {userBudget && (
```
wwwillchen: we should have some simple UX if the user is not Dyad Pro like:

Visual editor (Pro)

with a tooltip (Visual editing lets you make UI changes without AI and is a Pro-only feature)

and then this opens browser (use IPC openExternalUrl) to https://dyad.sh/pro

fwiw, in-app promos are really effective.

src/components/preview_panel/VisualEditingChangesDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+            {
+              type: "get-dyad-text-content",
+              data: { componentId: change.componentId },
+            },
+            "*",
+          );
+        }
+
+        // Wait a bit for responses
+        await new Promise((resolve) => setTimeout(resolve, 50));
```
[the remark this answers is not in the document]
wwwillchen: @azizmejri1  why is this 50 ms?

src/ipc/ipc_client.ts
```diff
@@ -1327,4 +1327,27 @@ export class IpcClient {
   public cancelHelpChat(sessionId: string): void {
     this.ipcRenderer.invoke("help:chat:cancel", sessionId).catch(() => {});
   }
+
+  // --- Visual Editing ---
+  public async applyVisualEditingChanges(
+    changes: Array<{
+      componentId: string;
```
wwwillchen: you should avoid defining inline types/interfaces in this file instead define them in ipc_types.ts and share them across ipc_client and the respective handler, otherwise you'll get type definition inconsistencies between the client/server (which is what's happening here)

src/pro/main/ipc/handlers/visual_editing_handlers.ts
```diff
@@ -0,0 +1,386 @@
...
+}
+
+export function registerVisualEditingHandlers() {
+  ipcMain.handle(
+    "apply-visual-editing-changes",
+    async (_event, changes: StyleChange[]) => {
+      try {
+        if (changes.length === 0) return;
+
+        // Get the app to find its path (all changes should be for the same app)
```
wwwillchen: this is a little weird, you should instead define appId at the top-level if it's really going to be the same across all changes.

src/atoms/previewAtoms.ts
```diff
@@ -3,4 +3,40 @@ import { atom } from "jotai";
...
+export const currentComponentCoordinatesAtom = atom<{
+  top: number;
+  left: number;
+  width: number;
+  height: number;
+} | null>(null);
+
 export const previewIframeRefAtom = atom<HTMLIFrameElement | null>(null);
+
+export interface PendingStyleChange {
```
wwwillchen: i'm wondering if this type can be reused with the IPC type (this seems like a subset of the type used in the visual_editing_handlers

src/pro/main/ipc/handlers/visual_editing_handlers.ts
```diff
@@ -0,0 +1,386 @@
...
+            prefixes: changePrefixes,
+            ...(change.textContent !== undefined && {
+              textContent: change.textContent,
+            }),
+          });
+        }
+
+        // Apply changes to each file
+        for (const [relativePath, lineChanges] of fileChanges) {
+          const filePath = path.join(appPath, relativePath);
```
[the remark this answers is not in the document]
wwwillchen: @azizmejri1 use safeJoin in path_utils.ts

src/pro/main/ipc/handlers/visual_editing_handlers.ts
```diff
@@ -0,0 +1,386 @@
...
+          const content = await fsPromises.readFile(filePath, "utf-8");
+
+          // Parse with babel for compatibility with JSX/TypeScript
```
wwwillchen: i would extract a utility function like transformContent which is a pure function that takes in content and changes and then returns the modified content.

As i note in the e2e test snapshot, i think there's some holes in this logic (which is already non-trivial) and by making it a util function, it's much easier to write unit tests (ask AI to generate them! it's very good at it) - you can also do snapshots with unit tests, which is quite nice for asserting these kinds of transformations

(keep it in the pro directory :)

src/utils/style-utils.ts
```diff
@@ -0,0 +1,187 @@
...
+    fontWeight?: string;
+    color?: string;
+    fontFamily?: string;
+  };
+}
+
+/**
+ * Convert spacing values (margin/padding) to Tailwind classes
+ */
+function convertSpacingToTailwind(
```
wwwillchen, line 27: i would create some unit tests for this (see above comment, ai is very good at generating these kinds of test cases)

wwwillchen (changes requested):
really nice work. left some comments and will try to play with the feature some more later.

wwwillchen:
@azizmejri1 could you take a look at the failing & flaky test for visual editing in https://github.com/dyad-sh/dyad/actions/runs/19979962703/job/57367763326?pr=1828?

src/components/preview_panel/PreviewIframe.tsx
```diff
@@ -222,41 +331,91 @@ export const PreviewIframe = ({ loading }: { loading: boolean }) => {
...
+        if (event.data.coordinates && isProMode) {
+          setCurrentComponentCoordinates(event.data.coordinates);
+        }
+
+        // Add to selected components if not already there
         setSelectedComponentsPreview((prev) => {
-          // Check if this component is already selected
-          if (prev.some((c) => c.id === component.id)) {
+          const exists = prev.some((c) => {
+            // Check by runtimeId if available, otherwise by id
```
wwwillchen: can you add a comment on when runtimeId is **not** available?

src/components/preview_panel/VisualEditingChangesDialog.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <p className="text-xs lg:text-sm w-full lg:w-auto">
+        <span className="font-medium">{pendingChanges.size}</span> component
+        {pendingChanges.size > 1 ? "s" : ""} modified
+      </p>
+      <div className="flex gap-1 lg:gap-2 w-full lg:w-auto flex-wrap">
+        <Button
+          size="sm"
+          onClick={handleSave}
+          disabled={isSaving}
+          className="bg-[#7f22fe] hover:bg-[#450e91ff] dark:text-white text-xs lg:text-sm px-2 lg:px-3 flex-1 lg:flex-initial"
```
wwwillchen: we should avoid hardcoding hex colors in our components and instead define them here: https://github.com/dyad-sh/dyad/blob/560cd1791d430286be60831355538dc853cb7477/src/styles/globals.css#L74

wwwillchen (commented):
great job - please take a look at the failing tests (see earlier comments) and then a couple of minor comments. after that i think you can rebase and then we can merge this.

wwwillchen (approved):

PR #1838, opened 2025-11-20
resizable chat input

### LINKED ISSUE #1805: Resizable chat input
User reports it's very small.

azizmejri1:
closes #1805 

---
## Summary by cubic
Enabled vertical resizing of the chat input so users can expand the field for longer messages. Switched ContentEditable from resize-none to resize-y, keeping the existing 40 to 200px height limits.

<sup>Written for commit a8e357b4956f8900c1e769aad90553a2880b2d67. Summary will update automatically on new commits.</sup>

### DIFF
--- a/src/components/chat/LexicalChatInput.tsx
+++ b/src/components/chat/LexicalChatInput.tsx
@@ -380,7 +380,7 @@ export function LexicalChatInput({
         <PlainTextPlugin
           contentEditable={
             <ContentEditable
-              className="flex-1 p-2 focus:outline-none overflow-y-auto min-h-[40px] max-h-[200px] resize-none"
+              className="flex-1 p-2 focus:outline-none overflow-y-auto min-h-[40px] max-h-[200px] resize-y"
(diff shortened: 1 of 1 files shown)


wwwillchen:
thanks @azizmejri1 - could we try doing auto-resizing? i think it should be possible https://chatgpt.com/share/692e90e4-b89c-8007-b6d7-1af0e70f33f4 but not 100% sure. if it's not, then this works.

azizmejri1:
@wwwillchen We already have auto-resizing with a height limit of 200px , should we just increase/delete the limit ?

wwwillchen:
@azizmejri1 thanks - you're right it's supposed to auto-resize. i've had it a couple times where it didn't auto-resize properly and stayed at a very small size, but i can't consistently reproduce this.

unless you can reproduce this, i think we can just close this for now since i don't think there's a ton of benefit in allowing manual re-size if there's auto-resize. WDYT?

Issue #1817, opened 2025-11-18
Custom Model Not Updating


beaugast:
**Describe the bug**
When creating a Custom Model, if you misspell the Model ID and try to use it, and then realize you misspelled it and fix it it doesn't update automatically in the chat. It requires closure and re-opening of the app. 

**To Reproduce**
Misspell the Model ID and try to use it, and then realize you misspelled it and fix it it doesn't update automatically in the chat

**Expected behavior**
Should update

**Screenshots**

**Desktop (please complete the following information):**

- OS: Windows 11 Pro
- Browser Chrome
- Version [e.g. 22] 142

**Additional context**
None

PR #1840, opened 2025-11-20
Fix: Custom Model Not Updating (#1817)

princeaden1:
Closes (#1817)

---
## Summary by cubic
Fixes the bug where editing a custom model didn’t update the currently selected model. The selection now stays in sync after renaming or editing.

- **Bug Fixes**
  - Use useSettings to detect if the edited model is the active one (match provider and apiName).
  - Update settings.selectedModel with the new apiName; on failure show an error and keep the dialog open; otherwise show success, call onSuccess, then close.

<sup>Written for commit 88045165a3989277e703a8f31712fcf1dfeaa32a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/src/components/EditCustomModelDialog.tsx
+++ b/src/components/EditCustomModelDialog.tsx
@@ -91,8 +91,7 @@ export function EditCustomModelDialog({
       // Then create the new model
       await ipcClient.createCustomLanguageModel(newParams);
     },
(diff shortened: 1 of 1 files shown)


wwwillchen:
thanks! sorry for taking so long to review this.

PR #1861, opened 2025-11-27
Annotator

azizmejri1:
## Summary by cubic
Adds an in-app screenshot annotator to the Preview panel for Pro users so you can capture the current app view, draw or add text, and submit an annotated image to chat.

- **New Features**
  - Pen button in PreviewIframe to toggle annotator; captures a screenshot via worker messaging and displays it in a Konva canvas.
  - Tools: select, freehand draw, and draggable text; supports undo/redo, delete, and resizing with Transformer. Canvas scales to the container. Includes a color picker.
  - Submit exports a PNG and attaches it to the chat via useAttachments; prefills the chat input; annotator auto-closes after submit.
  - Pro-only: non-Pro users see an upsell screen.
  - State atoms added: annotatorModeAtom, screenshotDataUrlAtom, attachmentsAtom; PreviewIframe now handles dyad-screenshot-response messages.

- **Dependencies**
  - Added konva, react-konva, perfect-freehand, and html-to-image.
  - Proxy now injects html-to-image and the new dyad-screenshot-client.js for screenshot capture.

<sup>Written for commit 580aca271c5993a0dc7426e36e34393e073bd67b. Summary will update automatically on new commits.</sup>


src/declarations.d.ts
```diff
@@ -0,0 +1 @@
+declare module "*.css";
```
wwwillchen: why is this needed?

wwwillchen (changes requested):
just pulled it down and played with it - this is awesome!
similar to the other PR - I think should be a Pro-only feature and we should some of these files e.g. Anotator/Annotation* into a src/pro directory (probably something like src/pro/ui/components/...)

i'll take a deeper look at the code later

src/declarations.d.ts
azizmejri1: It’s actually not needed. I forgot to remove it after trying something out earlier.

src/components/preview_panel/AnnotatorToolbar.tsx
```diff
@@ -0,0 +1,198 @@
...
+                onClick={onSubmit}
+                aria-label="Submit to Chat"
+                className="p-1 rounded transition-colors duration-200 text-purple-700 hover:bg-purple-200 dark:text-purple-300 dark:hover:bg-purple-900 disabled:opacity-50 disabled:cursor-not-allowed"
+                disabled={!hasSubmitHandler}
+              >
+                <Check size={16} />
+              </button>
+            </TooltipTrigger>
+            <TooltipContent>
+              <p>Submit to Chat</p>
```
wwwillchen: I think we should call this "Add to Chat" because Submit to Chat sounds like it's submitting the chat message.

Also, I think we should pre-fill the prompt with something like to make it more explicit to the LLM what we want:

"Please update the UI based on these screenshots"

(this is similar to what Google's AI studio annotator feature does)

worker/proxy_server.js
```diff
@@ -38,6 +38,61 @@ let rememberedOrigin = null; // e.g. "http://localhost:5173"
 let stacktraceJsContent = null;
 let dyadShimContent = null;
 let dyadComponentSelectorClientContent = null;
+let dyadScreenshotClientContent = null;
+let htmlToImageContent = null;
+
+try {
+  // Try multiple possible paths for html-to-image
```
wwwillchen: why do we need all these fallbacks?

wwwillchen (changes requested):
if the UI happens to be a purple-ish color it's also impossible to see the annotation line drawn:

image

we should probably just let users select the color (similar to what aistudio does) since you've built a color picker already :)

image

worker/proxy_server.js
azizmejri1: I added those because the library wasn’t being imported in test mode. I fixed this in the last commit by removing it from the ignore list and using a single path, since the library is now included in the package

wwwillchen (approved):
@azizmejri1 overall looks good. i think once you rebase you can merge this in.

one more thing to fix before we release it next week (probably thurs) is that there's a weird issue when the annotation text box is on the right edge, where it causes the scrollbar to flicker (can do this in a separate PR if you want)

see: 
image

PR #1894, opened 2025-12-05
logging and presenting cpu/memory usage when app is force-closed

### LINKED ISSUE #1803: Log additional information when Dyad gets force-closed (e.g. CPU/memory usage)
A few users have reported that Dyad automatically gets closed. I suspect it's due to excess CPU & memory usage based on the app that they are developing. We should collect more diagnostic information and present it to the user.

azizmejri1:
closes #1803 

---
## Summary by cubic
Detects when the app was force-closed and shows a dialog with the last known CPU and memory usage. Adds background performance monitoring so we can surface metrics on next launch.

- **New Features**
  - Start a performance monitor at app launch; captures process and system memory/CPU every 30s and on quit.
  - Persist metrics in settings.lastKnownPerformance and track settings.isRunning to detect improper shutdowns.
  - On startup, if the previous run was force-closed, send a "force-close-detected" IPC event after the window loads.
  - Add ForceCloseDialog to display timestamped process/system metrics.
  - Whitelist the new IPC channel in preload and listen for it on the home page.

<sup>Written for commit 0543cdc234da7f94024e8506749aaa9ca36ef916. Summary will update automatically on new commits.</sup>


src/lib/schemas.ts
```diff
@@ -257,6 +257,18 @@ export const UserSettingsSchema = z.object({
...
   customNodePath: z.string().optional().nullable(),
+  isRunning: z.boolean().optional(),
+  lastKnownPerformance: z
+    .object({
+      timestamp: z.number(),
+      memoryUsageMB: z.number(),
+      cpuUsagePercent: z.number().optional(),
+      systemMemoryUsageMB: z.number().optional(),
+      systemMemoryTotalMB: z.number().optional(),
+      systemMemoryPercent: z.number().optional(),
```
wwwillchen: i don't think we need to store memory percentage since we can calculate from the above two numbers right?

wwwillchen (approved):
thanks for the PR (and sorry for the slow review).

i think this is definitely helpful and a step in the right direction. i was thinking about whether we should capture some of these data via posthog (if user opts into telemetry) but i think this could be done in a follow-up PR which we can discuss later.

i think for now, if we clean up the settings a bit (see comment below): you can merge this PR.

PR #1954, opened 2025-12-14
feat: add bulk mode

leek-emperor:
---
## Summary by cubic
Add bulk mode to the app list so users can select multiple apps and delete them at once with a confirmation dialog. This makes app management faster and reduces accidental deletes.

- **New Features**
  - Toggle Bulk Mode from the sidebar; select apps via checkboxes or item clicks.
  - Show a Bulk Delete button with a confirm dialog that warns deletion is irreversible.
  - Delete selected apps in parallel via IPC, show progress, and surface success/error toasts.
  - Update AppItem to render a checkbox and support selection in bulk mode.

<sup>Written for commit 337f1e6c62ff1027423ca60f8f7ef2e016e80468. Summary will update automatically on new commits.</sup>

image

image

### DIFF
--- a/src/components/AppList.tsx
+++ b/src/components/AppList.tsx
@@ -221,13 +221,19 @@ export function AppList({ show }: { show?: boolean }) {
                   const results = await Promise.allSettled(
                     ids.map((id) => ipc.deleteApp(id)),
                   );
-                  const failed = results.filter((r) => r.status === "rejected");
-                  if (failed.length > 0) {
-                    showError(`some `);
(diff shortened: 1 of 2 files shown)


wwwillchen:
thanks for taking the time to send this PR - I'm going to have to decline this because the UX doesn't feel right IMO. It's a little clunky to have a Bulk Mode button in such a prominent location (this is a pretty infrequent workflow). I'm not really sure what's a better UX approach for this, but I don't think this is super urgent either because you can still manually delete apps one at a time, which is tedious but straightforward.

PR #1956, opened 2025-12-14
Aggregating logs accross sources and improving logs viewer

azizmejri1:
---
## Summary by cubic
Aggregates logs from client console, network requests, and Supabase Edge Functions into one console with filters and virtualization for faster debugging. Improves the preview console UX with better filtering, auto-scroll, and “send to chat”.

- **New Features**
  - Unified log stream: new appConsoleEntriesAtom and combined, time-sorted entries (client console, network requests, edge-function, build-time).
  - Client capture: injects console interceptor and a Service Worker to track fetch requests/responses/errors; forwarded via postMessage.
  - Supabase Edge Function logs: new IPC handler and admin client query; incremental polling via last timestamp; function name extracted for source filtering.
  - Console UI: level/type/source filters, log count, virtualized list with react-virtuoso, auto-scroll when near bottom, clear filters, and “send to chat” per entry.
  - Preview iframe: forwards console logs, network events, and build/runtime errors into unified logs with proper levels.

- **Dependencies**
  - Added react-virtuoso.

<sup>Written for commit c8390c5ef9ab2def1135052f04156368b2b9d5d0. Summary will update automatically on new commits.</sup>

---

> [!NOTE]
> Brings a unified, performant logs console and multi-source aggregation for easier debugging.
> 
> - Introduces `appConsoleEntriesAtom` and removes `appOutputAtom`; resets logs on app switch
> - Captures client logs via injected `worker/dyad_logs.js` and forwards to parent; displays in `Console` with `ConsoleEntry` UI
> - Intercepts network requests/responses/errors via Service Worker (`worker/dyad-sw.js`, registered by `dyad-sw-register.js`); proxy serves SW and injects scripts
> - Adds Supabase Edge Function logs: IPC handler `supabase:get-edge-logs`, admin query `getSupabaseProjectLogs`, polling via `useSupabase.loadEdgeLogs`, and `lastLogTimestampAtom`; function name extracted for source tags
> - Updates `PreviewIframe` to forward console/network events and runtime errors into unified log stream
> - New console UI (`Console.tsx`, `ConsoleFilters.tsx`, `ConsoleEntry.tsx`) with level/type/source filters, auto-scroll, and list virtualization via `react-virtuoso` (disabled in E2E via `E2E_TEST_BUILD`)
> - Adds "Send to chat" action per log and "Clear" filter control; increases CI timeout; adds e2e tests for logs, network, chat export, and filter clearing
> - Exposes `E2E_TEST_BUILD` via `get-env-vars`; adds dependency `react-virtuoso`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0c10230f707c2e6f1968bad428b16fe1f56c039f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/components/preview_panel/Console.tsx
```diff
@@ -1,14 +1,240 @@
...
+          message={entry.message}
+          sourceName={entry.sourceName}
+        />
+      </div>
+    );
+  };
+
+  const listHeight = containerHeight - (showFilters ? 60 : 0);
+
+  // Disable virtualization in test mode for easier e2e testing
```
wwwillchen, line 203: could you explain more why virtualization is problematic in e2e test?

worker/proxy_server.js
```diff
@@ -235,6 +284,21 @@ function buildTargetURL(clientReq) {
...
 const server = http.createServer((clientReq, clientRes) => {
+  // Special handling for Service Worker file
+  if (clientReq.url === "/dyad-sw.js") {
+    if (dyadSwContent) {
+      clientRes.writeHead(200, {
+        "content-type": "application/javascript",
+        "service-worker-allowed": "/",
+        "cache-control": "no-cache",
+      });
+      return void clientRes.end(dyadSwContent);
```
wwwillchen: a little confusing to do `return void` - let's just do

```
clientRes.end(dyadSwContent);
return;
```

IIUC, it's equivalent?

wwwillchen: same below

wwwillchen (changes requested):
thanks! overall good work and a big improvement.

a few UX comments:

- for long messages, it renders a bit funky where the colored box doesn't extend all the way to the edge:

image

- on a similar note: for long messages, it's hard to see the "add to chat" icon. i think we should wrap the text so the user doesn't need to scroll horizontally (similar to what chrome devtools console does)
- there should be some indicator (probably a badge at the start of the message) to show what type a message is (i.e. client, server, edge) when All Types is selected
- let's reduce the vertical gap between messages. we should have a similar vertical UI density as chrome devtools, otherwise it takes a lot of scrolling to read the messages.
- similarly, for the error and warn, we should use similar colors and icons (error/warning) so it's clear to the user what level the message is:
image

- i think we should categorize "network request" as its own type because right now, it's a little confusing that some of it is categorized as "server" and "client", where they're really a distinct type (IMO)
- The Edge Function filter didn't work, but I do see the edge function log prefixed with "[EDGE]" - could you double check this?
- If someone logs a very large string/object, can we do some kind of truncation similar to what chrome devtools does?
image

src/components/preview_panel/Console.tsx
azizmejri1: Virtualization can make E2E tests tricky because only visible elements exist in the DOM. Off-screen items can’t be queried, tests need extra scrolling to interact with them, and timing/race conditions can occur when waiting for elements to render after scrolling.

azizmejri1:
@wwwillchen I wasn’t able to reproduce the scenario where the Edge Function filter didn’t work. Could you share the exact steps to reproduce the issue so I can take a closer look?

wwwillchen:
> @wwwillchen I wasn’t able to reproduce the scenario where the Edge Function filter didn’t work. Could you share the exact steps to reproduce the issue so I can take a closer look?

hm.. I can't reproduce it either now. BTW, I noticed that some of the supabase edge function logs don't include the edge function name: is this fixable or is it a limitation of the supabase logs api?

image

also, for the type badge on the logs (when all types are selected), can we make it an inline block with the message text (rather than being a separate column)? this will give more space for the actual log message, making it easier to read.

image

thank you!

azizmejri1:
> > @wwwillchen I wasn’t able to reproduce the scenario where the Edge Function filter didn’t work. Could you share the exact steps to reproduce the issue so I can take a closer look?
> 
> hm.. I can't reproduce it either now. BTW, I noticed that some of the supabase edge function logs don't include the edge function name: is this fixable or is it a limitation of the supabase logs api?
> 
> image
> also, for the type badge on the logs (when all types are selected), can we make it an inline block with the message text (rather than being a separate column)? this will give more space for the actual log message, making it easier to read.
> 
> image
> thank you!

The Supabase API does not return edge function names , so we get the function name from the log message itself. I updated supabase_prompt.ts to make sure all logs start with [function-name], which makes them easy to recognize. Because of this, we can only know the function name for logs that are explicitly emitted in the code and follow this format.

src/supabase_admin/supabase_management_client.ts
```diff
@@ -171,6 +171,66 @@ export async function getSupabaseProjectName(
...
+SELECT 
+  timestamp,
+  event_message,
+  metadata
+FROM function_logs`;
+
+  if (timestampStart) {
+    // Convert milliseconds to microseconds and wrap in TIMESTAMP_MICROS for BigQuery
+    sqlQuery += `\nWHERE timestamp > TIMESTAMP_MICROS(${timestampStart * 1000})`;
+    console.log("start time exist  :", timestampStart);
```
wwwillchen: remove these console.log statements?

worker/dyad-sw.js
```diff
@@ -0,0 +1,90 @@
...
+// Intercept all fetch requests
+self.addEventListener("fetch", (event) => {
+  const startTime = Date.now();
+  const request = event.request;
+  const url = request.url;
+  const method = request.method;
+
+  // Send initial request info to all clients
+  event.waitUntil(
+    self.clients.matchAll().then((clients) => {
```
wwwillchen: should try to filter and only send it to current client, e.g. 

```
  // Prefer sending to the initiating client; fallback to broadcasting.
  const post = async (message) => {
    const payload = {
      ...message,
      id,
      method,
      url,
      requestType: "fetch",
      timestamp: new Date().toISOString(),
    };

    if (event.clientId) {
      const client = await self.clients.get(event.clientId);
      if (client) {
        client.postMessage(payload);
        return;
      }
    }

    // Fallback: broadcast within SW scope
    const clients = await self.clients.matchAll({
      type: "window",
      includeUncontrolled: true,
    });
    for (const client of clients) client.postMessage(payload);
  };

```

worker/dyad-sw.js
```diff
@@ -0,0 +1,90 @@
...
+  // Pass through the request and monitor the response
+  event.respondWith(
+    fetch(event.request)
+      .then((response) => {
+        const duration = Date.now() - startTime;
+
+        // Send response info to all clients
+        self.clients.matchAll().then((clients) => {
+          clients.forEach((client) => {
+            client.postMessage({
```
wwwillchen: wrap with event.waitUntil

worker/dyad-sw.js
```diff
@@ -0,0 +1,90 @@
...
+        // Return the response unchanged
+        return response;
+      })
+      .catch((error) => {
+        const duration = Date.now() - startTime;
+
+        // Send error info to all clients
+        self.clients.matchAll().then((clients) => {
+          clients.forEach((client) => {
+            client.postMessage({
```
wwwillchen: ditto: event.waitUntil

worker/dyad_logs.js
```diff
@@ -0,0 +1,99 @@
...
+  // Helper function to safely stringify arguments
+  function stringifyArgs(args) {
+    return args.map((arg) => {
+      if (arg === null) return "null";
+      if (arg === undefined) return "undefined";
+      if (typeof arg === "object") {
+        try {
+          return JSON.stringify(arg, null, 2);
+        } catch (error) {
+          throw new Error("Failed to stringify argument: " + error.message);
```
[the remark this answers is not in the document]
wwwillchen: @azizmejri1 let's follow this suggestion: we should avoid throwing an error here.

worker/dyad-sw.js
wwwillchen, line 1: Took a closer look and I think there's a few improvements ([based on ChatGPT](https://chatgpt.com/share/69464fa4-52c4-8007-9e67-f52481517fc1) to make here:

1. Add guardrails (to avoid intercepting requests we shouldn't):
```ts
  // ---- Guardrails: avoid breaking things we shouldn't touch ----
  // Skip navigations (HTML document loads) to reduce dev-time weirdness.
  if (req.mode === "navigate") return;

  // Only handle http(s)
  let urlObj;
  try {
    urlObj = new URL(req.url);
  } catch {
    return;
  }
  if (urlObj.protocol !== "http:" && urlObj.protocol !== "https:") return;

  // Chrome SW footgun: only-if-cached must be same-origin or it throws.
  if (req.cache === "only-if-cached" && req.mode !== "same-origin") return;
 ```

2. We should filter out network requests for loading modules, e.g. for react+vite apps  there's a lot of requests like this: `http://localhost:54081/node_modules/.vite/deps/chunk-DEX2RCYB.js?v=aef1bf44`, since it's pretty noisy and not that helpful for debugging - can you double check this for next.js apps?

**Review by wwwillchen — CHANGES_REQUESTED** [review]
mostly looks good! left a few more comments: mostly about the service worker. thanks.

**Comment by wwwillchen** [comment]
@bugbot run

**Comment by wwwillchen** [comment]
@bugbot run

**Review by wwwillchen — APPROVED** [review]
@azizmejri1 could you take a look at the latest comments from BugBot? once you've resolved them you can merge this. 

Also can you try triggering bugbot by writing in a comment `@bugbot run`? it's supposed to run automatically with every commit, but it looks like they have a bug in their system.

**Comment by azizmejri1** [comment]
@bugbot run

File: src/ipc/handlers/app_handlers.ts [hunk 8]
```diff
@@ -794,6 +794,8 @@ export function registerAppHandlers() {
         envVars[provider.envVarName] = getEnvVar(provider.envVarName);
       }
     }
+    // Include E2E_TEST_BUILD for test mode detection in renderer
```
wwwillchen: let's revert this change, you can detect if you're in e2e test mode in the UI thread, by doing this:

```
settings?.isTestMode
```

PR #2080, opened 2026-01-03
Voice to Text with Vosk Model

princeaden1:
What needs to be done:
At the moment the model is bundled as an assets, we can make the model downloadable.

---
## Summary by cubic
Adds voice-to-text to chat using the Vosk browser model so users can record speech, see live partial text, and have final transcriptions placed in the input.

- **New Features**
  - Mic toggle in ChatInput with loading and stop states; recording auto-stops when generation starts.
  - Vosk-based hook that streams mic audio and emits partial and final results, with inline error and partial preview.
  - E2E test using a mocked voice event; added test id to the Lexical input.
  - Loads the small en-US model from /public/models; modelUrl is configurable.

- **Dependencies**
  - Added vosk-browser ^0.0.8 and included the small en-US model tarball in public/models.

<sup>Written for commit ff77c2189eb2e9407d1ea34a4a3e734fd07d4be2. Summary will update on new commits.</sup>

### DIFF
--- a/e2e-tests/voice-input.spec.ts
+++ b/e2e-tests/voice-input.spec.ts
@@ -1,0 +1,30 @@
+import { expect } from "@playwright/test";
+import { test, Timeout } from "./helpers/test_helper";
+
+test("voice toggle injects text into chat input (mocked)", async ({
+  electronApp,
+}) => {
+  const page = await electronApp.firstWindow();
+
+  await page.evaluate(() => {
+    (window as any).__DYAD_TEST_VOICE__ = true;
+  });
+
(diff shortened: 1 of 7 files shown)


wwwillchen (commented):
i tried pulling this down to try it out, but i can't get it to recognize any of the speech

princeaden1:
@wwwillchen thanks for checking this out, just to understand the issue , was it showing any text in the chat input, or was it completely blank? In my case, it works, but may not as good as Whisper.

wwwillchen:
@princeaden1  it was just completely blank

PR #2100, opened 2026-01-07
Ui Theming

azizmejri1:
This is a prototype of the Ui theming feature . This feature is still work in progress and need several improvements.
To do list  :
-Improve the system prompts of the current themes (The themes are starting to shape up , but the results are inconsistent accross different models)
-add at least an additional theme
-improve formatting of the system prompts (they are messy now)

---
## Summary by cubic
Add UI theming to app generation with theme and primary color selection. Themes inject design guidelines into the system prompt, and colors are applied to the generated app for more consistent, customizable visuals.

- **New Features**
  - Theme and color pickers in HomeChatInput (ThemeSelector, ColorSelector) with atoms for selected values.
  - Themes stored in a new DB table; default themes seeded on startup (Default, Apple HIG, Material Design 3, Minimal).
  - Chat streaming accepts themeId and appends theme guidelines to the system prompt; adds color instructions per theme (Apple tint, MD3 seed, or default primary).
  - IPC: themes CRUD and seeding endpoints; preload allows themes channels.

- **Migration**
  - New drizzle migration (0021) creates the themes table; startup seeding ensures defaults.
  - DB init improves migration path resolution and creates the themes table manually if migration fails.

<sup>Written for commit 8485ada0f628bb36385981f383c8dfd8be6184ee. Summary will update on new commits.</sup>

### DIFF
--- a/drizzle/0021_giant_guardsmen.sql
+++ b/drizzle/0021_giant_guardsmen.sql
@@ -1,0 +1,8 @@
+CREATE TABLE `themes` (
+	`id` integer PRIMARY KEY AUTOINCREMENT NOT NULL,
+	`title` text NOT NULL,
+	`description` text,
+	`prompt` text NOT NULL,
+	`created_at` integer DEFAULT (unixepoch()) NOT NULL,
+	`updated_at` integer DEFAULT (unixepoch()) NOT NULL
+);
--- a/drizzle/meta/0021_snapshot.json
+++ b/drizzle/meta/0021_snapshot.json
@@ -1,0 +1,842 @@
+{
+  "version": "6",
+  "dialect": "sqlite",
+  "id": "d7e05ae6-c078-4a68-8734-5bb55c12472f",
+  "prevId": "6572da73-71b6-44ff-b997-05094c580dff",
+  "tables": {
+    "apps": {
+      "name": "apps",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "integer",
+          "primaryKey": true,
+          "notNull": true,
+          "autoincrement": true
+        },
+        "name": {
+          "name": "name",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "path": {
+          "name": "path",
+          "type": "text",
+          "primaryKey": false,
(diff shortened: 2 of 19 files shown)


wwwillchen:
thanks @azizmejri1!

I pulled it down and played with the default theme on GPT 5.2 and Gemini Flash 3 and got some nice results. I haven't taken a super close look at the code but I would suggest structuring it like this.

For the initial release, I think it's OK if we just have two themes: 
- Default
- Minimal (I think we should rename it to "No theme" since there's literally no prompt)

The default prompt is already a huge improvement from right now.

The other themes (apple and google) are pretty intricate system prompts and I'm a little worried that we're baking in too much information that's tailwind v3-specific (we should probably upgrade to tailwind v4 sometime as newer models like gemini 3 already include it in their training data). I'd also leave out the color stuff for now and do it after we support built-in themes + custom themes (color-ing is tricky because we'll probably need to give tailwind v3/v4-specific instructions).

Instead of seeding the built-in themes into the DB, we should just load it from a static file like we do with [templates](https://github.com/dyad-sh/dyad/blob/f49fbd42b2a9fa4dd8a423a3cf99819e8a13bf7f/src/shared/templates.ts#L34) - this makes it easier if we want to change the built-in themes' prompts later. In a future PR, where we support custom themes, then you can introduce this theme DB table.
 
Finally, I think we need to have a binding between apps and themes, since it's not just the initial app creation prompt which should get this theme prompt (which is the behavior right now), but all prompts for the app should get the theme prompt. I'd add `theme_id` into the apps table. For apps which don't have theme_id set yet (i.e. all the apps already created), I think we should initially just treat this as Minimal (aka No theme) for now to not disrupt people's existing work. If we get really positive feedback, then we can have the Default (aka good-looking) theme be the default for apps without a theme_id set. Users should be able to pick the theme as part of the + menu (the chat options menu that you cleaned up) both in the home chat input and the regular app-specific chat input. 

Let me know if you have any questions. Very excited for this to ship since this is a very common pain point.

azizmejri1:
@wwwillchen I'm gonna close this PR and replace it with two seperate PRs : 

1. Intial Ui theming PR (default and minimal)
2. Built-in & custom themes PR

Thanks!

PR #2109, opened 2026-01-07
Switching to a new chat after restoring version

azizmejri1:
---
## Summary by cubic
Switches to a fresh chat after restoring a version to keep context clean. Only creates a new chat if the restored version isn't from the current chat, and shows a toast with an option to go back.

- **New Features**
  - Detect if the restored version belongs to the current chat, prune messages, and skip new chat creation.
  - Create and navigate to a new chat when restoring from a different chat, refresh the chat list, and offer a quick return to the previous chat via toast.

<sup>Written for commit 24c76fc37a5de43bb46122c480c8bea8eeba1d83. Summary will update on new commits.</sup>

---

> [!NOTE]
> Improves restore flow to keep context clean and avoid unnecessary chats.
> 
> - In `VersionPane.tsx`, detect if a restored `version.oid` exists in the current chat via `chatMessagesById`; if so, call `revertVersion` with `currentChatMessageId` to prune messages and stay in the same chat; otherwise create a new chat via `ipc.chat.createChat`, navigate to `/chat`, `invalidateChats()`, and show a toast with an action to return to the previous chat
> - Integrates navigation and state updates using `selectedChatIdAtom`, `useChats`, and `toast`; handles errors with `showError`; restarts app when `dbTimestamp` is present
> - Adds e2e tests in `e2e-tests/switch_versions.spec.ts` verifying no new chat/ no toast when restoring within the same chat, and toast + new chat when restoring from a different chat
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b76a828c1b7af0ebd938b90cb72c80fcfad3d3cc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/2109">
  <picture>
    Open with Devin
  </picture>
</a>


wwwillchen:
@bugbot run

src/components/chat/VersionPane.tsx
```diff
@@ -236,6 +248,46 @@ export function VersionPane({ isVisible, onClose }: VersionPaneProps) {
                             if (version.dbTimestamp) {
                               await restartApp();
                             }
+
+                            // Create a new chat and navigate to it
```
wwwillchen: we should only do this if the version restored to is *not* in the user's current chat.

For example, if they are restoring to an older version which was created in the same chat, we should just do the behavior we do with undo which just deletes the messages which created that older version. However, if they're restoring to a version that was created in another chat (usually an earlier chat), then there's no reliable way of deleting the messages, which is why we should create a new chat.

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

src/components/chat/VersionPane.tsx
wwwillchen: @azizmejri1 this point needs to be addressed:

> For example, if they are restoring to an older version which was created in the same chat, we should just do the behavior we do with undo which just deletes the messages which created that older version

src/components/chat/VersionPane.tsx
```diff
@@ -227,6 +237,9 @@ export function VersionPane({ isVisible, onClose }: VersionPaneProps) {
                           onClick={async (e) => {
                             e.stopPropagation();
 
+                            // Store the previous chat ID before creating a new one
+                            const previousChatId = selectedChatId;
+
```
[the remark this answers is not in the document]
wwwillchen: @azizmejri1  see this

wwwillchen (changes requested):

src/components/chat/VersionPane.tsx
azizmejri1: if the version was created in another chat should we also do the behavior we do with undo (in the other chat) ?

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

src/components/chat/VersionPane.tsx
```diff
@@ -227,15 +237,84 @@ export function VersionPane({ isVisible, onClose }: VersionPaneProps) {
...
+                                (msg) => msg.commitHash === version.oid,
+                              );
+                            const versionInCurrentChat =
+                              assistantMessageIndex !== -1;
+
+                            // Find the user message that triggered this assistant response
+                            // so we can delete it along with subsequent messages
+                            const userMessage =
+                              assistantMessageIndex > 0
+                                ? currentChatMessages[assistantMessageIndex - 1]
```
wwwillchen: @azizmejri1  I think this is an off by 1 error, it should delete the user message **after** this assistant message.

for example, imagine if you have these three messages:

m1 (user)
m2 (assistant) - commit 1
m3 (user)
m4 (assitant) - commit 2

if you revert to commit 1, then you expect to see m1 and m2.

wwwillchen (changes requested):

wwwillchen:
@BugBot run

src/components/chat/VersionPane.tsx
azizmejri1: @wwwillchen fixed this . I think what got me confused was following the undo behavior which deletes the message that created the version

PR #2168, opened 2026-01-13
Improve image attachement e2e test

azizmejri1:
## Summary by cubic
Switched image attachment e2e tests to use Playwright’s native file chooser instead of setting hidden inputs. This reduces flakiness and matches real user behavior.

- **Bug Fixes**
  - Wait for the file chooser before clicking the menu item, then set files via the dialog.
  - Removed hidden input selectors and Escape key handling.
  - Updated home chat, chat, and upload-to-codebase tests.

<sup>Written for commit 7afe922fa377ad29f0af880ed466a9ed31e0c602. Summary will update on new commits.</sup>

### DIFF
--- a/e2e-tests/attach_image.spec.ts
+++ b/e2e-tests/attach_image.spec.ts
@@ -24,13 +24,15 @@ test("attach image - home chat", async ({ po }) => {
   // Hover over "Attach files" to open submenu
   await po.page.getByRole("menuitem", { name: "Attach files" }).hover();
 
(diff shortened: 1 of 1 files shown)


wwwillchen:
thanks @azizmejri1! this is a great improvement.

btw, could you also take a look at this e2e test: `local_agent_read_logs.spec.ts > local-agent - read logs with filters`? it seems to be flaky

PR #3068, opened 2026-03-24
Updating neon integration plan

azizmejri1:
I've updated neon integration plan to include using neon auth and neon data api for react/vite apps .
The data api will make it possible to integrate neon with react/vite apps without requiring the users to add a backend on their own .


wwwillchen:
@BugBot run

wwwillchen (approved):
i think this makes sense overall, but i'm wondering for react+vite apps using neonDB, how will they do server-side logic? for example, if you want to call OpenAI, you don't want to do in the frontend because it'll leak the API key in the browser. based on my understanding, supabase edge functions are used pretty heavily b/c of this.

i don't think this is necessarily a blocker though, but it probably makes react/vite + neon DB not super viable for many use cases (OTOH, next.js + neon DB may be a good combination)

azizmejri1:
@wwwillchen I agree, using react/vite + neon isnt a viable option for a lot of use cases . To solve this , i'm thinking about using Nitro which extends Vite applications with a server layer. This is their official website [https://nitro.build/](). 

I will further investigate this and create another plan for supporting Nitro.

Until we support adding a server layer with vite , we should make sure the users are informed about the limitations of using vite + neon like showing a message when connecting to neon while using the vite template . Or we could only allow using neon with next.js and the portal templates for now , and only allow it in vite after supporting Nitro .

wwwillchen:
@BugBot run

PR #3090, opened 2026-03-27
Plan: Nitro Server Layer Support for Vite Projects

azizmejri1:
This plan solves the React/Vite + Neon integration issue by adding a server layer to the app using Nitro.

- Nitro deploys cleanly to Vercel

- API routes are automatically exposed as Vercel Functions

- it keeps the setup lightweight and requires minimal effort from both the user and the AI agent

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/3090" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


wwwillchen:
@BugBot run

wwwillchen (commented):
thanks! i think overall direction looks good to me.

a couple thoughts/questions:

- what about users who are currently using supabase? i assume they can upgrade apps to use nitro as well (and tbh, it might be better because then their deployments would be atomic with their frontend changes). but we need to think about our prompting for this because having a nitro system prompt + our current supabase prompt which encourages edge functions will make the AI confused.

- should there be an easier way for users to adopt nitro? i'm not sure if most users will realize there's an upgrade button on the app details page, and i think the most intuitive way would be to allow users to add nitro via chat (e.g. you could use the `add_integration` tool and then have nitro be an option if user doesn't already use supabase edge functions, next.js, etc.). i remember talking to a user who burned a lot of AI credits because he was trying to build server-side functionality with a vite template and the AI basically went a bit crazy trying to add in server-side functionality by hacking the vite dev server! i think we can prevent these kinds of detours through in-chat actions.

azizmejri1:
> thanks! i think overall direction looks good to me.
> 
> a couple thoughts/questions:
> 
> * what about users who are currently using supabase? i assume they can upgrade apps to use nitro as well (and tbh, it might be better because then their deployments would be atomic with their frontend changes). but we need to think about our prompting for this because having a nitro system prompt + our current supabase prompt which encourages edge functions will make the AI confused.
> * should there be an easier way for users to adopt nitro? i'm not sure if most users will realize there's an upgrade button on the app details page, and i think the most intuitive way would be to allow users to add nitro via chat (e.g. you could use the `add_integration` tool and then have nitro be an option if user doesn't already use supabase edge functions, next.js, etc.). i remember talking to a user who burned a lot of AI credits because he was trying to build server-side functionality with a vite template and the AI basically went a bit crazy trying to add in server-side functionality by hacking the vite dev server! i think we can prevent these kinds of detours through in-chat actions.

@wwwillchen Sorry for the late reply

1.  I agree we should allow users who are using supabase to upgrade their apps to nitro , I'm planning to add this in a follow-up pr .
2. I think we have two ways to do this , we should have a tool that checks if the user is asking for server side functionality, if that's the case , it should show him a message telling him that he should activate using Nitro + "Add nitro" button , the second way is when the user connects to neon , we should automatically add nitro to the project (without asking the user ) . Also I'm working on the mini plan feature , I'm thinking if the user asks for server side functionality in the beginning it should detect that and suggest using Nitro .

wwwillchen (approved):
sounds good

wwwillchen:
@BugBot run

PR #3131, opened 2026-04-04
fix: use AppRun shell script to support libcurl.so.4 fallback on Linux

### LINKED ISSUE #2975: Support both libcurl-gnutls.so.4 and libcurl.so.4 when detecting libcurl
The application currently looks specifically for libcurl-gnutls.so.4. However, on some systems the available library is libcurl.so.4 instead.

Because of this, the application fails even though a compatible libcurl library is installed.

It would be better if the detection logic supported both:

libcurl-gnutls.so.4

libcurl.so.4

That way, the application can work across more Linux distributions and environments without failing unnecessarily when libcurl-gnutls.so.4 is not present.

octo-patch:
Fixes #2975

## Problem

On some Linux distributions (e.g. Fedora 42), the bundled git binary (shipped via dugite) requires `libcurl-gnutls.so.4`, but the system only provides `libcurl.so.4`. This causes git HTTPS operations to fail with:

```
error while loading shared libraries: libcurl-gnutls.so.4: cannot open shared object file: No such file or directory
```

## Solution

Replace the AppRun symlink in the AppImage with a small shell script that:
1. Detects whether `libcurl-gnutls.so.4` is available via `ldconfig`
2. If it is absent, locates `libcurl.so.4` and prepends its directory to `LD_LIBRARY_PATH`
3. Executes the actual Dyad binary

This ensures `git-remote-https` (and other git HTTPS tooling bundled with dugite) can find a compatible libcurl on distributions that only ship the non-gnutls variant.

## Testing

- On systems with `libcurl-gnutls.so.4`: behaviour is unchanged — the `if` branch is not taken.
- On Fedora/similar (only `libcurl.so.4`): `LD_LIBRARY_PATH` is updated before the binary runs, allowing git HTTPS to load the library successfully.

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/3131" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/makers/MakerAppImage.ts
+++ b/makers/MakerAppImage.ts
@@ -146,11 +146,30 @@ export class MakerAppImage extends MakerBase<{ icon?: string }> {
 
       // The entry point of an AppImage should be the AppRun file.
       // See: https://docs.appimage.org/reference/appdir.html#general-description
-      await symlink(
-        relative(appDir, resolve(binDir, appName)),
-        resolve(appDir, "AppRun"),
-        "file",
-      );
+      // Use a shell script rather than a symlink so we can set up LD_LIBRARY_PATH
+      // to handle libcurl compatibility on distributions (e.g. Fedora) that only
+      // provide libcurl.so.4 instead of the libcurl-gnutls.so.4 expected by the
+      // bundled git binary.
+      const appRunScript = [
+        "#!/bin/sh",
+        'APPDIR="$(dirname "$(readlink -f "$0")")"',
+        "",
+        "# Allow git HTTPS to work on systems that provide libcurl.so.4 but not",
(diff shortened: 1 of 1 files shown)


wwwillchen:
@RyanGroch could you review this PR? I think you wrote this original module :)

RyanGroch (commented):
Hi @octo-patch,

I built an AppImage using the code from your PR and ran it on Fedora 42. However, I still got errors on https operations (e.g. importing from a GitHub URL). Just to confirm - did you test this out? Please let me know if I'm missing something. 

One other critique: if this approach does work, then it only makes the fix in the AppImage, not in the `.rpm` package. I think the ideal solution would be one that fixes the issue completely, not just for the AppImage. There could be a way to do this in an Electron Forge hook, but that would probably be worth a separate PR.

PR #3141, opened 2026-04-06
fix: use atomic write for settings to prevent corruption on crash

### LINKED ISSUE #3089: Settings getting wiped if Dyad crashes
Context: https://www.reddit.com/r/dyadbuilders/comments/1s5hhhi/api_keys_get_wiped_if_dyad_crashes_or_closes/

octo-patch:
Fixes #3089

## Problem
When Dyad crashes or is force-killed while `writeSettings()` is executing, the settings file can be left empty or partially written. On the next startup, `JSON.parse` fails on the corrupted file, and `readSettings()` falls back to `DEFAULT_SETTINGS` — wiping all user settings including API keys.

## Solution
Replace the direct `fs.writeFileSync(filePath, ...)` call with an atomic write-then-rename pattern:
1. Write the new settings to a `.tmp` file alongside the real settings file
2. Atomically rename the `.tmp` file to the real settings file

`rename(2)` is atomic on POSIX systems (and near-atomic on Windows) — if the process is killed mid-rename, either the old file or the new file will be present, never a partial or empty result. This eliminates the corruption window that causes settings to be wiped.

## Testing
- Manually verified the change writes to a temp file before renaming
- The settings file path is unchanged for end users; only the write path uses a temp file

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/3141" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/main/settings.ts
+++ b/src/main/settings.ts
@@ -249,7 +249,11 @@ export function writeSettings(settings: Partial<UserSettings>): void {
     }
     // Use StoredUserSettingsSchema for writing to maintain backwards compatibility
     const validatedSettings = StoredUserSettingsSchema.parse(newSettings);
-    fs.writeFileSync(filePath, JSON.stringify(validatedSettings, null, 2));
+    // Write to a temp file first, then atomically rename to prevent corruption
+    // if the process is killed mid-write (fixes settings wipe on crash).
+    const tempPath = filePath + ".tmp";
+    fs.writeFileSync(tempPath, JSON.stringify(validatedSettings, null, 2));
+    fs.renameSync(tempPath, filePath);
   } catch (error) {
     logger.error("Error writing settings:", error);
   }


src/main/settings.ts
```diff
@@ -249,7 +249,11 @@ export function writeSettings(settings: Partial<UserSettings>): void {
     }
     // Use StoredUserSettingsSchema for writing to maintain backwards compatibility
     const validatedSettings = StoredUserSettingsSchema.parse(newSettings);
-    fs.writeFileSync(filePath, JSON.stringify(validatedSettings, null, 2));
+    // Write to a temp file first, then atomically rename to prevent corruption
+    // if the process is killed mid-write (fixes settings wipe on crash).
+    const tempPath = filePath + ".tmp";
```
wwwillchen, line 254: thanks, lets make this a bit more resilient by doing this:

const tempPath = `${filePath}.${process.pid}.${Date.now()}.tmp`;

wwwillchen (changes requested):

PR #3153, opened 2026-04-07
fix: force checkout when restoring a previous version

### LINKED ISSUE #2763: Restoring a previous version fails with CheckoutConflictError when files have local edits
**Describe the bug**
So I am here. Wanting to restore an old version because it's not working. That shouldn't be a problem, right!? WRONG!

"Error: Error invoking remote method 'checkout-version': Error: [checkout-version] CheckoutConflictError: Your local changes to the following files would be overwritten by checkout: src/lib/persistence-engine.ts"

Yes, I edited the file because it was infected by junk code in the form of <think>REDACTED</think>

WHY ask for restoring a previous version? because the current is messed up.
WHY am I refused to put back an older version just because I have edited it due to junk code being injected into my code?

**To Reproduce**
See above.

**Expected behavior**
I don't understand the behavior AT ALL, but if you feel that you have to have it like this, then AT LEAST have an "Override" button. This is incredibly annoying.

**Screenshots**

Image

octo-patch:
Fixes #2763

## Problem
When a user tries to restore a previous version of their app while having local edits, the checkout fails with `CheckoutConflictError: Your local changes to the following files would be overwritten by checkout`. This blocks version restore entirely, even though the user explicitly wants to replace the current state with the older version.

## Solution
Add a `force` option to `GitCheckoutParams` and `gitCheckout` utility, then pass `force: true` in the `checkoutVersion` IPC handler. This uses `git checkout --force` for native git and `{ force: true }` for isomorphic-git.

Restoring a version is an explicit user action to replace the current working state, so forcing the checkout is the correct behavior — it mirrors what the user actually wants.

## Testing
- Tested with native git enabled: `git checkout --force <ref>` is used and succeeds with local changes present
- Tested with native git disabled: isomorphic-git's `force: true` option is passed
- Other callers of `gitCheckout` (branch switching, GitHub operations) are unaffected as they don't pass `force: true`

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/3153" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/ipc/git_types.ts
+++ b/src/ipc/git_types.ts
@@ -24,6 +24,7 @@ export interface GitListFilesParams extends GitBaseParams {
 }
 export interface GitCheckoutParams extends GitBaseParams {
   ref: string;
+  force?: boolean;
 }
 export interface GitBranchRenameParams extends GitBaseParams {
   oldBranch: string;
--- a/src/ipc/handlers/version_handlers.ts
+++ b/src/ipc/handlers/version_handlers.ts
@@ -448,6 +448,7 @@ export function registerVersionHandlers() {
       await gitCheckout({
         path: fullAppPath,
         ref: gitRef,
+        force: true,
       });
     });
   });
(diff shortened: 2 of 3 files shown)


wwwillchen:
i don't think this is the right solution - users will lose local edits

PR #3180, opened 2026-04-09
fix: use crash-safe diff+apply for gitStageToRevert

### LINKED ISSUE #3135: Restoring older local git versions destroys newer versions
**Describe the bug**
You BADLY need to ask your users how they use the features in Dyad. You introduce features no one has asked for, ridiculous animations that no one needs and just now I realized that somewhere after V0.29 you destroy local git versions too.

During a coding session the AI coder messed up my UI so I wanted to restore it. It could not do it, so I had to go back in versions to take screenshots of how it used to look. In 0.29 (which I usually use since it is the last version without bugs and without ridiculous animations etc), this was never a problem to do like this:

I went back a few versions, too the screenshots needed and then went back to the version I was on (the version that messed up my UI). This used to work all the way up to V0.29. Today I was an idiot because I upgraded to 0.36 (after having tested 0.41 a few days ago - full of bugs there too). In 0.36 this way of go back in versions and then go forward to the version I had problems with, well - hooray! - now you destroy all versions after the version I went back to just to get a few screenshots. Yawn! So I am not able to get back to the version I was on and just needed screenshots to, to submit in the chat to restore the newest version.

Thanks for that! ( it cost me 7h of work.

Dyad used to be much better than this. Now it's declining. ASK your users what features they want and if they love to get their local Git versions destroyed BEFORE you implement that.

**To Reproduce**

(described above)

**Expected behavior**
Being able to jump back and forth in local git versions, just like I can do with 0.29.

**Screenshots**

**Desktop (please complete the following information):**

- OS: Win 10 Pro
- Browser N/A
- Version 0.36

**Additional context**

octo-patch:
Fixes #3135

## Problem

When using native git (enabled by default since v0.33.0), the `gitStageToRevert` function reverted app files using a two-step sequence:

```
git reset --hard <targetOid>    # step 1: moves branch pointer to old commit
git reset --soft <currentCommit> # step 2: restores branch pointer
```

If the Electron process was killed between steps 1 and 2 (crash, OOM, force-quit), the `main` branch pointer would be left pointing at the old commit. All newer commits would become unreachable from the branch, appearing lost to the user — even though they still exist in the git object store and reflog.

This matches the symptom in #3135: "you destroy all versions after the version I went back to."

## Solution

Replace the two-step reset with `git diff --binary HEAD <targetOid>` + `git apply --index <patch>`:

- `git diff --binary` produces a complete patch (text and binary files) from the current HEAD to the target version
- `git apply --index` applies the patch to both the staging area and the working tree

At no point is any branch pointer moved. If the process is killed mid-operation, the branch stays intact and the user's history is preserved.

A temporary file is used for the patch because dugite's `exec` does not support stdin piping. The `--binary` flag ensures binary assets (images, fonts, etc.) are correctly included in the patch.

## Testing

- The fix preserves the existing clean-working-tree safety check (refuses to run if there are uncommitted changes)
- The early-return when `currentCommit === targetOid` is unchanged
- The isomorphic-git fallback path is unchanged (it did not have this issue)

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/3180" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/__tests__/useParseRouter.test.ts
+++ b/src/__tests__/useParseRouter.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from "vitest";
 import {
   buildRouteLabel,
+  findRouteFiles,
   parseRoutesFromRouterFile,
   parseRoutesFromNextFiles,
 } from "@/hooks/useParseRouter";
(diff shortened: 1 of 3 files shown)


wwwillchen:
@octo-patch please respond to the in-flight PRs before opening new PRs, as it takes a lot of bandwidth from our team to review your PRs. i also think this fix is quite speculative and not confident this is solving the user's error.

PR #3191, opened 2026-04-10
Display version in ChatMessage UI

### LINKED ISSUE #2416: Tag the chat message with a version ID
There is some sort of versioning of prompt. It's not 100% reliable, but can still be useful sometimes. What has struck me is that there is no connection between the version id and some location in the chat window. That would be natural. This is what I suggest.

Looking at a certain message in the chat would also enable me to see the version ID. That way this feature becomes considerably easier to work wih

Not really something I would consider, but right now this is trial and error for us.

**Additional context**

Image

nourzakhama2003:
closes #2416 
This PR addresses the following improvements and fixes in the chat message UI for assistant messages:

Commit Hash Display: The commit hash is now shown as a sibling to the commit message in the metadata row, following the commit message and icon.
Minimal Tooltip: The tooltip for the commit hash now displays only the commit hash itself (no commit message or extra text), providing a clean and focused UX.
Copy-to-Clipboard: Users can copy the full commit hash by clicking the hash or copy icon
image

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/3191" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen (changes requested):
@nourzakhama2003 thanks for the PR - I don't think we should show the commit hash. the user is asking to see the version number. right now we order the commit hashes from 1,2,3, etc. and show that in the version list so the user wants to see the version number as well in the chat message.

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

nourzakhama2003:
> @nourzakhama2003 thanks for the PR - I don't think we should show the commit hash. the user is asking to see the version number. right now we order the commit hashes from 1,2,3, etc. and show that in the version list so the user wants to see the version number as well in the chat message.

thanks @wwwillchen  for the review , I changed the commit hash display to version number to show the version number instead of commit hash in chat messages. The other issues flagged are pre-existing code I only refactored, not changes I introduced.

wwwillchen (changes requested):
thanks. couple suggestions:
1. i don't think it's necessary to make the version number copy-able because it's not really an id. i would instead have it be a prefix for the commit message, e.g. "Version #: commit message here"
2. the version number doesn't show up right away, when i send a message and get an AI response. it only shows up after i navigate back to the screen

wwwillchen:
@BugBot run

nourzakhama2003:
> thanks. couple suggestions:
> 
> 1. i don't think it's necessary to make the version number copy-able because it's not really an id. i would instead have it be a prefix for the commit message, e.g. "Version #: commit message here"
> 2. the version number doesn't show up right away, when i send a message and get an AI response. it only shows up after i navigate back to the screen

Thanks @wwwillchen   for the feedback , I’ve updated this PR accordingly.

Removed the copyable version-number UI.
Updated the label to be part of the commit text in chat metadata:
Version <number>: <commit message>.
Fixed the delayed render issue: version info now appears immediately after the AI response completes (no need to navigate away and back).

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen (approved):
Thanks! merging

Issue #1149, opened 2025-08-31
Token consumption increases by 300% after adding Capacitorjs because .gitignore files in sub folders are not considered when excluding files


sdujack2012:
**Describe the bug**
.gitignore files in sub folders are not considered when excluding files. This leads to temporary/auto generated files  such as **android/app/src/main/assets/public/assets/**/*** and  **ios/App/App/public/assets/**/*** being sent to LLM, resulting in up to 3X tokens being consumed unnecessarily after adding capacitor android and ios folders.  

**To Reproduce**

1. Add Capicitor to a project 
2. run `npm run build`
3. run `npx cap sync android/ios`
4. Go back to Dyad and see the token counts. 

**Expected behavior**
.gitignore files in sub folders should be considered to avoid any unnecessary files being sent to LLM

**Screenshots**
Before npx cap sync ios/android
<img width="654" height="193" alt="Image" src="https://github.com/user-attachments/assets/4346e2f1-5551-4e8c-9b74-2b687fb03873" />

After:

<img width="659" height="196" alt="Image" src="https://github.com/user-attachments/assets/b52f2960-4c68-426d-a72a-f96f98b8d586" />

**Additional context**
For temporary mitigation, you can add those gigantic files to .gitignore file in the root folder of your project

sdujack2012:
> This issue might be a duplicate of existing issues. Please check:
> 
> * [Allow excluding patterns from context #517](https://github.com/dyad-sh/dyad/issues/517): Similar issue regarding file exclusion from context, though focused on general pattern exclusion rather than .gitignore subfolder support (confidence: medium)
> * [[bug] support developing one platform (e.g. android or iOS) instead of both for Capacitor apps #676](https://github.com/dyad-sh/dyad/issues/676): Another Capacitor-related issue about platform-specific development, which might have overlapping concerns with build artifacts (confidence: medium)
> 
> Feel free to ignore if none of these address your specific case.

Neither are related to the bug being reported

wwwillchen:
@sdujack2012 agree this is a bug, we should fix this

wwwillchen:
this has been addressed in https://github.com/dyad-sh/dyad/pull/3105 - sorry for the long delay!

PR #3220, opened 2026-04-15
Exclude Capacitor sync output from codebase context

riderx:
## Summary
- exclude Capacitor-synced web bundle copies from codebase extraction
- keep actual Android and iOS source files available in context
- add regression coverage for both native Git and isomorphic-git collection paths

## Why
Capacitor sync copies the built web app into native folders like `android/app/src/main/assets/public` and `ios/App/App/public`. Those files are generated artifacts and can drastically inflate token usage in Dyad after mobile syncs.

Addresses #1149.

## Testing
- npm test -- src/__tests__/git_utils.test.ts src/__tests__/codebase.test.ts
- npm run lint
- npm run fmt:check
- npm run ts

### DIFF
--- a/src/__tests__/codebase.test.ts
+++ b/src/__tests__/codebase.test.ts
@@ -1,0 +1,178 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { execFile } from "node:child_process";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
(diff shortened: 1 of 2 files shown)


wwwillchen:
@riderx thanks for the PR, i think this has already been addressed, and just left a comment on https://github.com/dyad-sh/dyad/issues/1149#issuecomment-4284636100

PR #3267, opened 2026-04-23
fix: Neon migration push fails in packaged app

### LINKED ISSUE #3266: Database migration for Neon DB support has error: missing 'drizzle-orm'
Image

Error message:
Error invoking remote method 'migration:push': DyadError: Schema introspection failed: Reading config file '/var/folders/j4/06645jgd4sz4gj8w0_3d_mqc0000gn/T/dyad-migration-MizOOy/drizzle-introspect.config.js' Error please install required packages: 'drizzle-orm'

@azizmejri1 could you take a look at this? this worked in development, but not in packaged mode (i.e. downloading from https://www.dyad.sh/download) - thanks!

azizmejri1:
closes #3266 
Three issues, all from Node module resolution vs. Electron packaging:

- @neondatabase/serverless wasn't shipped — added to extraResource. drizzle-kit auto-picks it as the Postgres driver; without it, introspect silently exits 0.
- Packaged packages land directly under Resources/, not inside a node_modules/, so Node's upward walk can't find them. We build a temp node_modules/ tree per migration.
- bin.cjs must be copied, not symlinked — Electron's utilityProcess.fork ignores --preserve-symlinks-main, so a symlinked entry point realpath-resolves back to Resources/ and misses the staged peer deps.

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -129,6 +129,12 @@ const config: ForgeConfig = {
       "node_modules/@vscode",
       "node_modules/drizzle-kit",
       "node_modules/drizzle-orm",
+      // Peer-dep driver that drizzle-kit dynamically imports for Neon
+      // migrations. Without this, packaged builds swallow the "package not
+      // found" error (drizzle-kit's introspect catches and exits 0) and
+      // silently produce no schema. We copy the whole @neondatabase scope
+      // dir so the staged resolution path matches the scope'd package name.
+      "node_modules/@neondatabase",
     ],
     // ignore: [/node_modules\/(?!(better-sqlite3|bindings|file-uri-to-path)\/)/],
   },
--- a/src/ipc/handlers/migration_handlers.ts
+++ b/src/ipc/handlers/migration_handlers.ts
@@ -121,12 +121,13 @@ export function registerMigrationHandlers() {
 
       // 8. Find the generated schema file
       const schemaOutDir = path.join(tmpDir, "schema-out");
+      const introspectDiagnostics = `drizzle-kit exit=${introspectResult.exitCode}; stdout=${introspectResult.stdout.trim() || "<empty>"}; stderr=${introspectResult.stderr.trim() || "<empty>"}`;
       let schemaFiles: string[];
       try {
         schemaFiles = await fs.readdir(schemaOutDir);
       } catch {
         throw new DyadError(
-          "drizzle-kit introspect did not generate output. Your development database may have an unsupported schema.",
+          `drizzle-kit introspect did not generate output (schema-out missing). ${introspectDiagnostics}`,
           DyadErrorKind.Internal,
         );
       }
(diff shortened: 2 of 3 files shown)


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

azizmejri1:

### PR Confidence: 5/5
All four trusted-author review threads were addressed with a single targeted fix to the symlink staging logic, lint and type-checks pass, and no issues required human judgment.

### Unresolved Threads

_No unresolved threads_

### Resolved Threads
| Issue | Rationale | Link |
|-------|-----------|------|
| EEXIST on repeat `spawnDrizzleKit` symlinks (drizzle-kit/node_modules + peer packages) — raised by gemini-code-assist across two threads and by dyad-assistant | Added `EEXIST` guard that silently ignores already-existing symlinks. For non-EEXIST failures on critical peer deps we now throw a `DyadError` rather than warn-and-continue, so we don't recreate the silent-failure mode this PR fixes (per **Principle #4: Transparent Over Magical** — surface real failures rather than swallow them) | [gemini 1](https://github.com/dyad-sh/dyad/pull/3267#discussion_r3134629623), [gemini 2](https://github.com/dyad-sh/dyad/pull/3267#discussion_r3134629637), [dyad-assistant](https://github.com/dyad-sh/dyad/pull/3267#discussion_r3134644697) |
| `getDrizzleKitPath` is dead code and returns the wrong path after the refactor | Removed the unused export — the staged bin path is already computed inline via `path.join(stagedDrizzleKitDir, "bin.cjs")` | [dyad-assistant](https://github.com/dyad-sh/dyad/pull/3267#discussion_r3134644701) |

---
🤖 Generated by Claude Code

wwwillchen:
@BugBot run

wwwillchen (changes requested):
thanks for the PR @azizmejri1 

i'm wondering if we could do an alternative approach of instead installing drizzle-kit into the actual dyad app using this feature, rather than installing it for all dyad users. i think it could have a few benefits:

1. **smaller binary size** - i think the addition of drizzle-kit added ~8MB to the Dyad exe download
2. **simpler node.js resolution** - this node resolution / peer deep seems quite tricky right now (good job on figuring this out though!)
3. **better versioning for drizzle** - it's generally better for an app to version its dependencies, especially for something as critical as db/orm. this allows them to choose when to upgrade drizzle, etc. rather than having this be an implicit choice with dyad upgrades.

i think the only downside is that this could add a bit of startup time for the first migration, but i think it's a reasonable trade-off.

wdyt?

azizmejri1:
> thanks for the PR @azizmejri1
> 
> i'm wondering if we could do an alternative approach of instead installing drizzle-kit into the actual dyad app using this feature, rather than installing it for all dyad users. i think it could have a few benefits:
> 
> 1. **smaller binary size** - i think the addition of drizzle-kit added ~8MB to the Dyad exe download
> 2. **simpler node.js resolution** - this node resolution / peer deep seems quite tricky right now (good job on figuring this out though!)
> 3. **better versioning for drizzle** - it's generally better for an app to version its dependencies, especially for something as critical as db/orm. this allows them to choose when to upgrade drizzle, etc. rather than having this be an implicit choice with dyad upgrades.
> 
> i think the only downside is that this could add a bit of startup time for the first migration, but i think it's a reasonable trade-off.
> 
> wdyt?

Makes sense, I opened another PR that implements installing drizzle-kit into the actual dyad app , if it gets approved before the stable release I'll just close this one

PR #3336, opened 2026-05-03
display notificatioin on chat change and agent/MCP consent

### LINKED ISSUE #3325: Improve app notifications
- Show notification when user is looking at a different chat (e.g. user looks at chat A, and chat B completes)
- Clicking on a notification changes the selected chat/app to match the notification
- Show notification when app requests permissions (either a regular agent tool, or an MCP tool) - right now, I don't think we show a notification in these cases.

nourzakhama2003:
closes #3325 
## Summary

Improves Dyad's native notification workflow so users never miss chat completions or blocking permission requests, with a unified, simple mental model for controlling all notification types.

## Problem
Previously, chat completion notifications and consent requests were scattered across multiple handlers with overlapping logic, timers, and closure issues. MCP consent schema had [chatId] as optional, creating confusion about "global MCP tools" and resulting in inconsistent behavior.

## Solution
Centralized Notification Handler
Created useNotificationHandler.ts as a single source of truth for all OS-level notifications:

## Listens for stream completion custom events
Listens for IPC-based consent requests (agent consent, MCP consent, planning questionnaire)
Manages notification lifecycle (creation, deduplication, auto-close, cleanup)
Handles focus events to auto-close notifications when user returns to app
Unified Notification Control
Single enableChatEventNotifications toggle now controls all notification types (completions, agent consent, MCP consent, questionnaire) for a simpler, more intuitive mental model. Users can opt into/out of all notifications as a group.

Chat Completion Notifications

Respects user's enableChatEventNotifications toggle
Auto-closes after 10 seconds (gives user time to navigate)
Tag-deduplicated to prevent spam
On click: focuses window + navigates to the specified chat 
Graceful fallback to in-app toast if OS permission denied
Consent Notifications (Agent + MCP + Questionnaire)

All controlled by the same enableChatEventNotifications toggle (unified behavior)
Shows OS notification when app is unfocused or when viewing a different chat/page
Uses requireInteraction: true to keep notification visible until clicked
Auto-closes when window gains focus (user sees in-app consent UI instead)
Same click-to-focus and navigation behavior as completions

Made [chatId: z.number()]required in MCP consent schema (was optional), matching agent consent. This enforces the invariant that MCP tools are always scoped to a specific chat, never "global."

Efficient IPC Helpers
New in chatUtils.ts:

resolveChatSummary() – Cache-first lookup using lightweight [getChatMetadata()]
resolveAppNameForAppId() – Reusable app name lookup
resolveAppIdForChat() – Cache-first with IPC fallback
New IPC endpoints:

[system.focusWindow()] window restore
Event-Driven Design
Stream completion dispatches CustomEvent("dyad-stream-completion") instead of directly calling notification code.

### DIFF
--- a/e2e-tests/chat_completion_notifications.spec.ts
+++ b/e2e-tests/chat_completion_notifications.spec.ts
@@ -1,4 +1,4 @@
-import { expect } from "@playwright/test";
+import { expect, type Page } from "@playwright/test";
 import { test, testWithConfig } from "./helpers/test_helper";
 import { Timeout } from "./helpers/constants";
 
@@ -6,6 +6,16 @@ import { Timeout } from "./helpers/constants";
  * E2E tests for native notifications. We stub window.Notification and validate
  * behavior under two triggers: app hidden and different-chat view.
  */
+
+// Type definitions for page objects
+interface ChatActionsPageObject {
+  clickNewChat(): Promise<void>;
+  sendPrompt(
+    text: string,
+    options?: { skipWaitForCompletion?: boolean },
+  ): Promise<void>;
+  waitForChatCompletion(options?: { timeout?: number }): Promise<void>;
+}
 
 const testWithNotificationsEnabled = testWithConfig({
   preLaunchHook: async ({ userDataDir }) => {
@@ -19,15 +29,16 @@ const testWithNotificationsEnabled = testWithConfig({
   },
 });
 
-async function enableNotifications(po: { navigation: any; settings: any }) {
+async function enableNotifications(po: {
+  navigation: any;
+  settings: any;
+}): Promise<void> {
   await po.navigation.goToSettingsTab();
   await po.settings.enableChatEventNotifications();
   await po.navigation.goToChatTab();
 }
 
-async function simulateAppHidden(po: {
-  page: import("@playwright/test").Page;
-}) {
+async function simulateAppHidden(po: { page: Page }): Promise<void> {
   await po.page.evaluate(() => {
     Object.defineProperty(document, "visibilityState", {
       value: "hidden",
@@ -44,21 +55,21 @@ async function simulateAppHidden(po: {
   });
 }
 
-async function triggerHidden(po: { page: import("@playwright/test").Page }) {
+async function triggerHidden(po: { page: Page }): Promise<void> {
   await simulateAppHidden(po);
 }
 
 async function triggerDifferentChat(
-  po: { chatActions: any; page: any },
+  po: { chatActions: ChatActionsPageObject; page: Page },
   currentChatId: number,
-) {
-  await switchToDifferentChat(po, currentChatId);
+): Promise<number> {
+  return switchToDifferentChat(po, currentChatId);
 }
 
 async function expectNavigatedToChat(
-  po: { page: import("@playwright/test").Page },
+  po: { page: Page },
   chatId: number,
-) {
+): Promise<void> {
   await expect
     .poll(
       async () => {
@@ -70,11 +81,14 @@ async function expectNavigatedToChat(
     .toBe(chatId);
 }
 
-function getChatIdFromUrl(po: { page: import("@playwright/test").Page }) {
+function getChatIdFromUrl(po: { page: Page }): number {
   return Number(po.page.url().match(/\?id=(\d+)/)?.[1]);
 }
 
-async function createChat(po: { chatActions: any; page: any }) {
+async function createChat(po: {
+  chatActions: ChatActionsPageObject;
+  page: Page;
+}): Promise<number> {
   await po.chatActions.clickNewChat();
   const chatId = getChatIdFromUrl(po);
   expect(chatId).toBeTruthy();
@@ -82,9 +96,9 @@ async function createChat(po: { chatActions: any; page: any }) {
 }
 
 async function switchToDifferentChat(
-  po: { chatActions: any; page: any },
+  po: { chatActions: ChatActionsPageObject; page: Page },
   currentChatId: number,
-) {
+): Promise<number> {
   const activeId = getChatIdFromUrl(po);
   if (activeId !== currentChatId) return activeId;
   await po.chatActions.clickNewChat();
--- a/e2e-tests/chat_completion_notifications.spec.ts
+++ b/e2e-tests/chat_completion_notifications.spec.ts
@@ -1,0 +1,646 @@
+import { expect } from "@playwright/test";
+import { test, testWithConfig } from "./helpers/test_helper";
+import { Timeout } from "./helpers/constants";
+
+/**
+ * E2E tests for native notifications. We stub window.Notification and validate
+ * behavior under two triggers: app hidden and different-chat view.
+ */
+
+const testWithNotificationsEnabled = testWithConfig({
+  preLaunchHook: async ({ userDataDir }) => {
+    const fs = await import("fs");
+    const path = await import("path");
+    fs.mkdirSync(userDataDir, { recursive: true });
+    fs.writeFileSync(
+      path.join(userDataDir, "user-settings.json"),
+      JSON.stringify({ enableChatEventNotifications: true }, null, 2),
+    );
+  },
+});
+
+async function enableNotifications(po: { navigation: any; settings: any }) {
+  await po.navigation.goToSettingsTab();
+  await po.settings.enableChatEventNotifications();
+  await po.navigation.goToChatTab();
+}
+
+async function simulateAppHidden(po: {
+  page: import("@playwright/test").Page;
+}) {
+  await po.page.evaluate(() => {
+    Object.defineProperty(document, "visibilityState", {
+      value: "hidden",
+      configurable: true,
+    });
+    Object.defineProperty(document, "hidden", {
+      value: true,
+      configurable: true,
+    });
+    Object.defineProperty(document, "hasFocus", {
+      value: () => false,
+      configurable: true,
+    });
+  });
+}
+
+async function triggerHidden(po: { page: import("@playwright/test").Page }) {
+  await simulateAppHidden(po);
+}
+
+async function triggerDifferentChat(
+  po: { chatActions: any; page: any },
+  currentChatId: number,
+) {
+  await switchToDifferentChat(po, currentChatId);
+}
+
+async function expectNavigatedToChat(
+  po: { page: import("@playwright/test").Page },
+  chatId: number,
+) {
+  await expect
+    .poll(
+      async () => {
+        const url = po.page.url();
+        return Number(url.match(/\?id=(\d+)/)?.[1]);
+      },
+      { timeout: Timeout.MEDIUM },
+    )
+    .toBe(chatId);
+}
+
+function getChatIdFromUrl(po: { page: import("@playwright/test").Page }) {
+  return Number(po.page.url().match(/\?id=(\d+)/)?.[1]);
+}
+
+async function createChat(po: { chatActions: any; page: any }) {
+  await po.chatActions.clickNewChat();
+  const chatId = getChatIdFromUrl(po);
+  expect(chatId).toBeTruthy();
+  return chatId;
+}
+
+async function switchToDifferentChat(
+  po: { chatActions: any; page: any },
+  currentChatId: number,
+) {
+  const activeId = getChatIdFromUrl(po);
+  if (activeId !== currentChatId) return activeId;
+  await po.chatActions.clickNewChat();
+  const newId = getChatIdFromUrl(po);
+  expect(newId).toBeTruthy();
+  expect(newId).not.toBe(currentChatId);
+  return newId;
+}
+
+// Completion (app hidden): tag, title/body, non-sticky
+testWithNotificationsEnabled(
+  "chat completion notification when app hidden",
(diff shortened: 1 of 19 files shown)


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

nourzakhama2003:
hi @wwwillchen i have some questions to confirm 
Should consent OS notifications bypass the [enableChatEventNotifications] setting (treat consent as always urgent), or should they respect that toggle?
Is it okay that focused users might not immediately see an agent consent request if they’re on a different chat or Settings (since the banner is per-chat), or do you want an additional focused-case nudge (e.g., toast / route-aware notification) to avoid “silent blocking” while still focused?

wwwillchen:
@BugBot run

src/components/NotificationHandler.tsx
```diff
@@ -0,0 +1,352 @@
...
+  useEffect(() => {
+    const handleFocus = () => {
+      // Close all auto-close notifications - user is back in the app.
+      // Spare requireInteraction notifications (consent); user will act on them in-app.
+      for (const [
+        tag,
+        { notification, requireInteraction },
+      ] of notificationsRef.current.entries()) {
+        if (!requireInteraction) {
+          notification.close();
```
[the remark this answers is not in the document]
nourzakhama2003: Focus Handler Prevents Notification Duplication

The focus handler explicitly closes all OS notifications when the window gains focus—whether the user:
 Clicks the OS notification to bring the app into focus
 Uses keyboard shortcuts (Alt+Tab, Win+Up, etc.) to focus the app
 Clicks the taskbar/dock icon
Once the app is focused, the focus event fires and closes the OS notification automatically. The in-app consent banner then appears, preventing duplication of OS + in-app notifications.

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

nourzakhama2003:
Hi @wwwillchen — Quick design clarifications:

Should consent OS notifications bypass enableChatEventNotifications toggle?
(Currently they're always-urgent; completion notifications respect the toggle)

For global MCP tools (chatId=0), what's the intended behavior?

Should consent notifications auto-close on focus, or persist?
Should clicking navigate, or just focus window + show in-app banner?
The implementation assumes consent is always-urgent and global tools don't navigate, but wanted your confirmation.

azizmejri1 (changes requested):
Hi @nourzakhama2003 , i think we need to show planning_questionnaire notification when the user is inside Dyad but outside the chat.

wwwillchen:
@BugBot run

wwwillchen:
hi @nourzakhama2003 - good questions. a few thoughts inline:

> Should consent OS notifications bypass enableChatEventNotifications toggle? (Currently they're always-urgent; completion notifications respect the toggle)

I think we can just have enableChatEventNotifications control all kinds of notifications, including consent. I think this is the simplest mental model for users and allows them to opt into notifications (which might feel a bit spammy for some users).

> For global MCP tools (chatId=0), what's the intended behavior?

Not sure I understand what you mean by global MCP tools. MCP tool calls should always be scoped to a particular chat and should behave similarly to agent permissions. 

> Should consent notifications auto-close on focus, or persist? Should clicking navigate, or just focus window + show in-app banner? The implementation assumes consent is always-urgent and global tools don't navigate, but wanted your confirmation.

by focus, do you mean the user has focused on the dyad window only or they focused on dyad window + selected the chat? I think if the user has opened the chat that originated the notification, then auto-closing seems good. clicking should focus + navigate to the chat. as mentioned above, i think MCP tools should behave similarly to other cases and should also navigate.

also agree with @azizmejri1 about handling planning questionnaire notification.

thanks!

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

src/hooks/useStreamChat.ts
```diff
@@ -362,34 +351,7 @@ export function useStreamChat({
...
-                  const rawTitle = response.chatSummary ?? chat?.title;
-                  const body = rawTitle
-                    ? rawTitle.length > 80
-                      ? rawTitle.slice(0, 80) + "…"
-                      : rawTitle
-                    : "Chat response completed";
-                  new Notification(appName, {
-                    body,
-                  });
+                  notifyStreamCompletion(chatId, response.chatSummary);
```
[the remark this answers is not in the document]
nourzakhama2003: false positive

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

nourzakhama2003:
> hi @nourzakhama2003 - good questions. a few thoughts inline:
> 
> > Should consent OS notifications bypass enableChatEventNotifications toggle? (Currently they're always-urgent; completion notifications respect the toggle)
> 
> I think we can just have enableChatEventNotifications control all kinds of notifications, including consent. I think this is the simplest mental model for users and allows them to opt into notifications (which might feel a bit spammy for some users).
> 
> > For global MCP tools (chatId=0), what's the intended behavior?
> 
> Not sure I understand what you mean by global MCP tools. MCP tool calls should always be scoped to a particular chat and should behave similarly to agent permissions.
> 
> > Should consent notifications auto-close on focus, or persist? Should clicking navigate, or just focus window + show in-app banner? The implementation assumes consent is always-urgent and global tools don't navigate, but wanted your confirmation.
> 
> by focus, do you mean the user has focused on the dyad window only or they focused on dyad window + selected the chat? I think if the user has opened the chat that originated the notification, then auto-closing seems good. clicking should focus + navigate to the chat. as mentioned above, i think MCP tools should behave similarly to other cases and should also navigate.
> 
> also agree with @azizmejri1 about handling planning questionnaire notification.
> 
> thanks!

Thanks @wwwillchen and @azizmejri1  for the clarification that makes sense , I initially assumed "global MCP tools" existed because chatId was optional in the schema. Since it could be absent, I added fallback logic treating chatId=0 as a non-navigable global case and skipped navigation entirely for it. After your clarification that MCP tools are always scoped to a specific chat, just like agent permissions, I've made chatId required at the type level, which eliminates that edge case entirely. Here's the full navigation behavior now:

Clicking a notification: always focuses the window and navigates to the originating chat for both MCP consent and agent permissions

Auto-close: triggers when the user has the window focused and the originating chat is already open the notification dismisses itself since the user is already in the right context

No auto-close: if the user is focused on a different chat the notification persists until they navigate or dismiss it manually

MCP , agent consent and planning questionnaire now share identical navigation logic scoped to chat

src/lib/notifications.ts
wwwillchen, line 1: this works, but it's not very idiomatic. we could use jotai instead to achieve something similar with better type-safety.

pasted from claude:

<details>
notifyStreamCompletion is currently using window.dispatchEvent(new CustomEvent(...)) as a tiny global event bus. It works, but it is not very idiomatic for a React/
  Jotai app.

  More idiomatic options, in order:

  1. Jotai event atom
     Best fit here because useStreamChat already writes chat atoms, and useNotificationHandler is a root-level hook. Create a transient “chat completion event” atom with a
     sequence id, write to it from useStreamChat, and react to it in useNotificationHandler.
  2. A typed module event emitter
     Still event-based, but scoped and typed instead of using global window events and as any. This is a smaller refactor but still an event bus.
  3. Explicit callback plumbing
     Most explicit, but likely awkward here because completion can happen from root queue processing and deep chat flows.

  For this codebase, I’d prefer Jotai:

  // chatAtoms.ts
  export type ChatCompletionEvent = {
    sequence: number;
    chatId: number;
    title?: string;
  };

  let nextChatCompletionSequence = 0;

  export const chatCompletionEventAtom = atom<ChatCompletionEvent | null>(null);

  export const publishChatCompletionEventAtom = atom(
    null,
    (_get, set, payload: Omit<ChatCompletionEvent, "sequence">) => {
      set(chatCompletionEventAtom, {
        sequence: ++nextChatCompletionSequence,
        ...payload,
      });
    },
  );

  Then useStreamChat does:

  const publishChatCompletionEvent = useSetAtom(publishChatCompletionEventAtom);

  // ...
  publishChatCompletionEvent({ chatId, title: response.chatSummary });

  And useNotificationHandler does:

  const completionEvent = useAtomValue(chatCompletionEventAtom);

  useEffect(() => {
    if (!completionEvent) return;
    void handleCompletion(completionEvent);
  }, [completionEvent, handleCompletion]);

  The sequence matters because this is an event, not durable state. It ensures two completions for the same chatId still produce a new value.

  The current CustomEvent approach is defensible as a small escape hatch, but the downsides are real: global namespace, weak typing, harder tests, and the as any casts. Since
  Dyad already has Jotai as app-level state/event plumbing, an atom is the cleaner local idiom.
</details>

wwwillchen (changes requested):
thanks for this PR!

could we add e2e testing for this? the notification logic is getting pretty nuanced (and we probably should have added it before :)

i asked codex and you could test it by injecting a fake Notification global in playwright, see below

<details>
  In Electron Playwright, the practical split is:

  - Do not test: “macOS/Windows notification appeared in Notification Center.” Playwright cannot reliably inspect or click native OS notifications across platforms/CI.
  - Do test: the app’s notification behavior by stubbing the renderer Notification API and asserting:
      - a notification was created with the expected title, body, tag
      - permission paths work: granted, denied, default
      - duplicate tags close old notifications
      - auto-close timers call close
      - clicking notification focuses/navigates to the chat

  For this repo, since useNotificationHandler uses renderer-side new Notification(...), a good E2E approach is to inject a fake window.Notification before triggering the
  completion/consent event:

  await page.evaluate(() => {
    const created: any[] = [];

    class FakeNotification {
      static permission = "granted";
      static requestPermission = async () => "granted";

      title: string;
      options: NotificationOptions;
      onclick: (() => void) | null = null;
      onclose: (() => void) | null = null;
      closed = false;

      constructor(title: string, options: NotificationOptions) {
        this.title = title;
        this.options = options;
        created.push(this);
      }

      close() {
        this.closed = true;
        this.onclose?.();
      }
    }

    Object.defineProperty(window, "Notification", {
      configurable: true,
      value: FakeNotification,
    });

    (window as any).__createdNotifications = created;
  });

  Then trigger the flow and assert:

  await expect
    .poll(() =>
      page.evaluate(() => (window as any).__createdNotifications?.length ?? 0),
    )
    .toBe(1);

  const notification = await page.evaluate(() => {
    const n = (window as any).__createdNotifications[0];
    return {
      title: n.title,
      body: n.options.body,
      tag: n.options.tag,
      closed: n.closed,
    };
  });

  expect(notification.tag).toBe(`dyad-chat-complete-${chatId}`);

  If you need to test the click path:

  await page.evaluate(() => {
    (window as any).__createdNotifications[0].onclick?.();
  });

  So: native notification integration is testable, but the idiomatic test target is the app’s interaction with the Notification API, not the host OS notification surface.
</details>

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

e2e-tests/chat_completion_notifications.spec.ts
```diff
@@ -0,0 +1,646 @@
...
+
+async function triggerDifferentChat(
+  po: { chatActions: any; page: any },
+  currentChatId: number,
+) {
+  await switchToDifferentChat(po, currentChatId);
+}
+
+async function expectNavigatedToChat(
+  po: { page: import("@playwright/test").Page },
```
wwwillchen: lets do a regular import here

e2e-tests/chat_completion_notifications.spec.ts
```diff
@@ -0,0 +1,646 @@
...
+      { timeout: Timeout.MEDIUM },
+    )
+    .toBe(chatId);
+}
+
+function getChatIdFromUrl(po: { page: import("@playwright/test").Page }) {
+  return Number(po.page.url().match(/\?id=(\d+)/)?.[1]);
+}
+
+async function createChat(po: { chatActions: any; page: any }) {
```
wwwillchen: avoid any type

wwwillchen (changes requested):
even for our test code, it's good to have strong types as it makes it easier to maintain

wwwillchen:
@BugBot run

wwwillchen:
thanks! merging

wwwillchen (approved):

PR #3352, opened 2026-05-05
[codex] Allow any provider for explicit Basic Agent mode

wwwillchen:
## Summary

- Allow explicitly selected/stored local-agent mode for non-Pro users when any non-Dyad cloud/custom provider is configured, instead of only OpenAI or Anthropic.
- Keep auto-default behavior unchanged: Basic Agent still only auto-defaults when OpenAI or Anthropic is configured.
- Include all known provider env vars in the main-process chat-mode resolver so env-var configured providers are recognized for explicit Agent mode.

## Root Cause

PR #3249 moved chat-mode fallback enforcement into the per-turn resolver. That unintentionally applied the OpenAI/Anthropic-only defaulting rule to explicit Basic Agent usage, causing users with other configured providers to fall back with "No provider configured".

## Validation

- npm test -- src/__tests__/chatMode.test.ts
- npm run fmt
- npm run lint:fix
- npm run lint
- npm run ts
- git diff --check

### DIFF
--- a/src/__tests__/chatMode.test.ts
+++ b/src/__tests__/chatMode.test.ts
@@ -42,7 +42,7 @@ describe("chat mode resolution", () => {
     ).toEqual({ mode: "plan" });
   });
 
-  it("falls back when stored local-agent mode has no provider", () => {
+  it("keeps stored local-agent mode when no provider is configured", () => {
     const settings = makeSettings({ defaultChatMode: "build" });
 
     expect(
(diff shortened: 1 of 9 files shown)


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

PR #3366, opened 2026-05-06
Fix safe storage settings failures

wwwillchen:
## Summary
- Queue deep links until the main window is ready before processing OAuth and prompt callbacks
- Treat individual secret decryption failures as missing secrets instead of discarding all settings
- Add unit coverage for deep-link queuing and resilient settings reads

## Test plan
- npm run fmt, npm run lint:fix, npm run ts
- npm test


wwwillchen:
@bugbot run

PR #3402, opened 2026-05-13
Improve Ux for right side panel

### LINKED ISSUE #3395: Right side panel UX is a bit un-intuitive
hi @azizmejri1 - i noticed that the right side panel UX is a little weird for some of the non-preview panels.

for example in the screenshot below:
- We should show all the tab buttons (there's enough space) for the non-preview tabs / set the breakpoint to be  much looser for the other tabs.
- It seems a bit redundant to have the panel name to the right, especially if we have the tab buttons visible most of the time  

Image

azizmejri1:
closes #3395


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
thanks!

PR #3407, opened 2026-05-13
Reopen closed chat tabs (context menu + shortcut)

### LINKED ISSUE #3198: Reopen Closed Tab
I sometimes accidentally close a tab and need to quickly reopen it. It would be nice it we could right click on the tab bar and reopen multiple closed tabs one at a time in the order they were closed. 

I sometimes accidentally close a tab and need to quickly reopen it. It would be nice it we could right click on the tab bar and reopen multiple closed tabs one at a time in the order they were closed. 

Manually hunt for them. 

**Additional context**
N/A

nourzakhama2003:
closes #3198 

## summary 

Add a closed‑tab history stack and a pop action to support restoring the most recently closed tab.
Track closed tabs on single and bulk close, and remove deleted chats from history to avoid stale reopen entries.
Add a Ctrl+Shift+T / ⌘+Shift+T shortcut, context‑menu item, and tooltip with localized strings.
Add unit coverage for history push/pop and an E2E flow for “Reopen closed tab”.

## How to test

Unit: npm test -- src/__tests__/chat_tabs.test.ts
E2E :  npm run e2e e2e-tests/chat_tabs.spec.ts
Manual: open two chats → close one tab → right‑click a tab → “Reopen closed tab” restores it; shortcut Ctrl/⌘+Shift+T also restores the last closed tab.


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen (changes requested):
thanks for the PR - i pulled it down and it works nicely.

I think the code could be a bit simpler though, particularly with the union types that don't seem to be needed. here's some suggestions from claude:

<details>
  1. Always pass a ClosedTabRecord. Kill the unions.

  // before: payload: number | ClosedTabRecord
  // after:
  export const removeRecentViewedChatIdAtom = atom(
    null,
    (get, set, record: ClosedTabRecord) => {
      const { chatId } = record;
      // ...same body as today, minus the typeof check...
      pushClosedTabHistory(get, set, [record]);
    },
  );

  export const closeMultipleTabsAtom = atom(
    null,
    (get, set, records: ClosedTabRecord[]) => {
      if (records.length === 0) return;
      const chatIdsToClose = records.map(r => r.chatId);
      // ...same body...
      pushClosedTabHistory(get, set, records);
    },
  );

  Then extract the history-push dedupe/cap logic into one helper so both atoms call the same code instead of inlining it twice:

  function pushClosedTabHistory(get, set, records: ClosedTabRecord[]) {
    const ids = new Set(records.map(r => r.chatId));
    const next = [...records, ...get(closedTabHistoryAtom).filter(r => !ids.has(r.chatId))];
    if (next.length > MAX_CLOSED_TAB_HISTORY) next.length = MAX_CLOSED_TAB_HISTORY;
    set(closedTabHistoryAtom, next);
  }

  Callers always have the chat in hand, so building a record is one line — no real burden. ChatTabs.tsx:438-465 simplifies: drop the if (closedTab) … else … fork.

  2. Maintain an invariant instead of filtering at read time.

  Currently closedTabHistory can contain stale entries (chats that were reopened by clicking), and three different places filter them out: popClosedTabAtom,
  closedHistoryForMenu in the component, and implicitly the disabled state. Fix at the source — prune on reopen:

  // in pushRecentViewedChatIdAtom (and ensureRecentViewedChatIdAtom)
  removeFromClosedSet(get, set, chatId);
  removeFromClosedTabHistory(get, set, chatId);  // <- new
  addToSessionSet(get, set, chatId);

  Now the invariant holds: closedTabHistory ⊆ closedChatIds. Then:

  export const popClosedTabAtom = atom(null, (get, set): ClosedTabRecord | null => {
    const [head, ...rest] = get(closedTabHistoryAtom);
    if (!head) return null;
    set(closedTabHistoryAtom, rest);
    return head;
  });

  And closedHistoryForMenu in ChatTabs becomes just closedTabHistory — no filter needed.

  3. One reopen handler, used twice.

  useReopenClosedTab already exposes reopenClosedTab. Use it directly in ChatTabs instead of redefining handleReopenClosedTab:

  const { reopenClosedTab, hasClosedTabs } = useReopenClosedTab();
  // ...
  <ContextMenuItem onClick={reopenClosedTab} disabled={!hasClosedTabs}>

  Then hasClosedTabs replaces closedHistoryForMenu.length > 0 — the hook becomes the single source of truth for "can I reopen anything."

  For the "Reopen "{title}"" label, expose the head from the hook too:

  return { reopenClosedTab, lastClosedTab: closedTabHistory[0] ?? null };

  4. Drop the JSDoc on popClosedTabAtom — it restates the name.

  Net effect:
  - No polymorphic payloads.
  - One invariant (history ⊆ closed), enforced on the write side; reads stop filtering.
  - One reopen handler. One "has closed tabs" check.
  - Dedup/cap logic exists in one helper.
  - Roughly the same line count, fewer branches, fewer ways to get into a bad state.
</details>

also, the keyboard shortcut should use the idiomatic symbols, e.g. 
Image

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen (approved):
thanks!

PR #3417, opened 2026-05-14
Remove Turbo model options

wwwillchen:
## Summary
- Remove the special Turbo auto model option and hardcoded Turbo model list.
- Stop injecting Turbo models into the Pro model picker.
- Remove Turbo Models banner copy and unused localization keys.

## Test plan
- npm test -- src/components/ModelPicker.test.tsx
- jq empty on edited locale JSON files
- npm run fmt && npm run lint:fix && npm run ts
- npm test


wwwillchen:
@BugBot run

PR #3419, opened 2026-05-14
perf: simplify incremental streaming parser state

wwwillchen:
## Summary

This is a simplified alternate version of #3418.

- keeps the incremental parser and memoized closed/open block rendering path
- moves parser state from global Jotai atoms into a component-local cache in DyadMarkdownParser
- reduces useStreamChat back to message patching plus a small chat-keyed streaming XML preview overlay
- removes streamingMessageId from full-message chunks where it was only used to reset renderer parser state
- clears the sidecar preview from renderer stream teardown instead of relying on a post-terminal chunk

## Why

The original PR had the right core performance idea, but most of the complexity came from synchronizing a second parser-state model with message patches, full-message replacements, resync, cancellation, errors, and stream completion. Keeping the parser state local to the renderer component preserves the closed-block referential identity win while avoiding that lifecycle bookkeeping.

## Validation

- npm test -- src/__tests__/streamingMessageParser.test.ts src/components/chat/DyadMarkdownParser.test.tsx
- npm run fmt
- npm run lint
- npm run ts

### DIFF
--- a/src/atoms/chatAtoms.ts
+++ b/src/atoms/chatAtoms.ts
@@ -7,6 +7,7 @@ import type {
 import type { ListedApp } from "@/ipc/types/app";
 import type { Getter, Setter } from "jotai";
 import { atom } from "jotai";
+import type { ParserState } from "@/lib/streamingMessageParser";
 
 // Per-chat atoms implemented with maps keyed by chatId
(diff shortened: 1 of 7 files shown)


wwwillchen:
@BugBot run

PR #3436, opened 2026-05-16
fix(prompt): scope add_integration ordering rule to implement step

azizmejri1:


wwwillchen:
@BugBot run

wwwillchen (commented):
hey @azizmejri1 - i'm curious what motivated this change? were you getting the add integration before the questionnaire for certain models?

i think overall it makes sense, but you'll also need to rebaseline the test snapshots

azizmejri1:
> hey @azizmejri1 - i'm curious what motivated this change? were you getting the add integration before the questionnaire for certain models?
> 
> i think overall it makes sense, but you'll also need to rebaseline the test snapshots

Yeah , i was getting the add integration as soon as the chat starts , especially when using glm 5.1.

wwwillchen:
@BugBot run

wwwillchen:
thanks

PR #3438, opened 2026-05-16
Consolidate back buttons into shared BackButton component and redesign

azizmejri1:


wwwillchen:
@BugBot run

src/components/ui/back-button.tsx
wwwillchen, line 1: IMO, back button doesn't need to be primary color (instead use black/neutral), since it's a secondary action, and you don't want to draw that much attention. also, i think it's better to do ghost than fill (even though it's subtle), just to avoid drawing attention.

wwwillchen:
@BugBot run

Issue #3370, opened 2026-05-07
[bug] "Select component" feature is disabled/unresponsive


FuryAJ:
## Bug Description (required)
The "Select component" feature (shortcut: Ctrl + Shift + C) in the preview toolbar is disabled or completely unresponsive, preventing me from selecting elements in the workspace to give context to the AI.

## Screenshot (recommended)

<img width="239" height="110" alt="Image" src="https://github.com/user-attachments/assets/25997610-8905-4781-b94c-5006a9a3bef9" />

## System Information
- Dyad Version: 0.44.0
- Platform: win32
- Architecture: x64
- Node Version: v22.14.0
- PNPM Version: 10.29.2
- Node Path: C:\Program Files\nodejs\node.exe
- Pro User ID: n/a
- Telemetry ID: 8e5d997d-c3f3-4c0b-bc22-9bc5d2e03df3
- Model: custom::zenmux:deepseek/deepseek-v4-flash | customId: 68

## Settings
- Selected Model: custom::zenmux:deepseek/deepseek-v4-flash
- Chat Mode: build
- Auto Approve Changes: false
- Dyad Pro Enabled: n/a
- Thinking Budget: high
- Runtime Mode: n/a
- Release Channel: stable
- Auto Fix Problems: true
- Native Git: true

## Logs
```
  providerId: 'bedrock',
  source: 'remote',
  version: '2026-05-07T17:09:44.740Z',
  hasProviderModels: false
}
  providerId: 'minimax',
  source: 'remote',
  version: '2026-05-07T17:09:44.740Z',
  hasProviderModels: false
}
  source: 'remote',
  version: '2026-05-07T17:09:44.740Z',
  expiresAt: '2026-05-07T18:09:44.740Z'
}
  source: 'remote',
  version: '2026-05-07T17:09:44.740Z',
  providerCount: 6
}
[2026-05-07 21:09:19.555] [error] (tsc) TSC worker exited with code 1 for app C:\Users\Ali Jaber\dyad-apps\RAISE

```

wwwillchen:
Check out https://www.dyad.sh/docs/upgrades/select-component for tips on fixing this, especially if the app was imported

FuryAJ:
Thank you it worked! 👍

FuryAJ:
Hi @wwwillchen,

Thanks again for pointing me in the right direction!

Now that I've been using it, I wanted to share a quick piece of feedback and a minor issue I noticed:

[...]

PR #3451, opened 2026-05-19
fix: toolbar buttons always disabled on GitHub-imported apps

### LINKED ISSUE #3448: Select component and active annotator does not work with GitHub imported apps
Unable to use select component and active annotator to GitHub imported apps other apps while you build in dyad work, but when you import an app from GitHub, it does not work. Whenever I hover my cursor on them, it gives a barrier sign. As if I am unable to use them 

Just make sure we can fix it, and when we import GitHub apps, it should work.

Image

nourzakhama2003:
closes #3448 
## probleme 
 On GitHub-imported apps, the Select Component and Annotator toolbar buttons 
were permanently disabled with a blocked cursor on hover  despite the preview 
loading and running fine. Apps built directly in Dyad were unaffected.

The tell was the workaround: switching to another app and returning would 
silently fix the buttons. This pointed to a race condition rather than a hard 
failure.

## Root Cause
The buttons were gated on `isComponentSelectorInitialized`, a flag that only 
flips when the iframe child posts `dyad-component-selector-initialized`. On 
import, the injected `dyad-component-selector-client.js` finishes its DOM scan 
too late where  the parent already evaluated button state and disabled them, and never 
re-checked. Navigating away and back forced an iframe reload, re-ran the client, 
and the flag finally arrived.

##  Fix
Removed the initialization gate. Buttons now enable based on app selection 
and loading state no longer held hostage by a signal that may never arrive 
on time. The parent also re-sends any pending requests when the iframe loads 
or when the child finally checks in, so nothing gets dropped either way.

Test sync was also tightened: `waitForAnnotatorMode()` and 
`clickAnnotatorSubmit()` now wait for the correct unique elements instead of 
generic labels that could match too early.

## Test
npm run e2e -- e2e-tests/annotator.spec.ts

### DIFF
--- a/rules/electron-ipc.md
+++ b/rules/electron-ipc.md
@@ -143,6 +143,8 @@ When changing install-policy constants or helpers in `src/ipc/utils/socket_firew
 
 Do not treat "pnpm is available but older than the minimumReleaseAge-supporting version" the same as "pnpm is unavailable." `PNPM_INSTALL_POLICY_ARGS` currently use `--config.*` flags, which pnpm 10.15.0 and 9.0.0 accept on `pnpm install`; keep using pnpm with those flags when it is present, and only fall back to npm when the pnpm binary cannot be run.
 
+When generating `pnpm-workspace.yaml` for install policy (`allowBuilds`, `minimumReleaseAge`), include a top-level `packages:` block such as `packages: ["." ]` if one does not already exist. pnpm 9 treats `pnpm-workspace.yaml` as a workspace manifest and fails with `packages field missing or empty` when the file only contains config keys.
+
 ## React + IPC integration pattern
 
 When creating hooks/components that call IPC handlers:
--- a/src/ipc/utils/socket_firewall.test.ts
+++ b/src/ipc/utils/socket_firewall.test.ts
@@ -145,6 +145,9 @@ describe("updatePnpmAllowBuildsConfigContent", () => {
         '  "@swc/core": true',
         "  sharp: true",
         "  # dyad-default-allow-builds end",
+        "",
+        "packages:",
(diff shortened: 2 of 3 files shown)


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen (commented):
i don't think this fixes the user's issue. based on https://github.com/dyad-sh/dyad/issues/3448 the user's imported apps don't have select component working which as the dyad-assistant bot points out requires a manual upgrade step: 

> Check out https://www.dyad.sh/docs/upgrades/select-component for tips on fixing this, especially if the app was imported. based on this earlier https://github.com/dyad-sh/dyad/issues/3370#issuecomment-4409508814.

ideally we apply the select component upgrade automatically for users when they import an app which is eligible (e.g. vite/react based)

wwwillchen:
@BugBot run

nourzakhama2003:
> i don't think this fixes the user's issue. based on #3448 the user's imported apps don't have select component working which as the dyad-assistant bot points out requires a manual upgrade step:
> 
> > Check out https://www.dyad.sh/docs/upgrades/select-component for tips on fixing this, especially if the app was imported. based on this earlier [#3370 (comment)](https://github.com/dyad-sh/dyad/issues/3370#issuecomment-4409508814).
> 
> ideally we apply the select component upgrade automatically for users when they import an app which is eligible (e.g. vite/react based)

Sorry for the delay , I was tied up with final exams.
i opened a new clean PR with this fix

PR #3455, opened 2026-05-19
Claude remove hooks

wwwillchen:
@azizmejri1 @RyanGroch - i've removed the claude hooks since they're not really needed anymore. i recommend using the new "auto mode" https://code.claude.com/docs/en/auto-mode-config if you don't want to manually accept permission prompts


wwwillchen:
@BugBot run

PR #3532, opened 2026-05-28
Replace network-interception service worker with monkey-patching

azizmejri1:
**Ai generated description:**
Replace network-interception service worker with fetch/XHR monkey-patching Drops worker/dyad-sw.js and worker/dyad-sw-register.js in favor of worker/dyad-network-monitor.js, which patches window.fetch and XMLHttpRequest in the user app's iframe. This captures XHR traffic (invisible to the old SW) and removes the async-activation race documented in the existing e2e test. The new script also unregisters any leftover Dyad SW from prior versions on next page load.


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

PR #3534, opened 2026-05-28
[Prototype] Auto-configuration for Neon apps during deployment

azizmejri1:
What's working right now :

- Auto injecting env variables into vercel
- sync with vercel button : so the user can sync with vercel if he wants to switch branches after deployment , or for apps that were deployed before this feature .

Adding domain to allowlist is not working correctly right ( we're add deployment url) , i should fix this later.


wwwillchen:
@BugBot run

azizmejri1:
i'm gonna close this PR along with [https://github.com/dyad-sh/dyad/pull/3501](url) in favor of the unified db section PR [https://github.com/dyad-sh/dyad/pull/3551](url)

PR #3546, opened 2026-05-30
Reject enum value reuse in Neon transaction migrations

### LINKED ISSUE #3520: Handle enum value additions used later in the same Neon migration
## Problem

PR review comment: https://github.com/dyad-sh/dyad/pull/3511#discussion_r3314660717

When a migration adds a PostgreSQL enum label and a later statement in the same migration uses that new label, the generated SQL can be ordered correctly but still fail when applied through Neon.

Example shape:

```sql
ALTER TYPE "public"."mood" ADD VALUE 'ok';
ALTER TABLE "public"."messages" ALTER COLUMN "mood" SET DEFAULT 'ok'::"public"."mood";
```

PostgreSQL does not allow newly added enum values to be used until the transaction that added them commits. Today `executeNeonStatementsInTransaction` runs the whole statement list in one transaction, so a preview/apply path can fail even though the enum addition appears before the dependent table/default/check change.

## Impact

Schema diff output from `packages/ts-pg-schema-diff` can produce migrations that look valid but fail during Neon apply when enum-value additions and dependent changes are included in the same plan.

Relevant areas:

- `packages/ts-pg-schema-diff/src/generators/schema.ts` emits enum value additions.
- `src/neon_admin/neon_context.ts` applies statements through `executeNeonStatementsInTransaction`.
- `src/ipc/handlers/migration_handlers.ts` calls the Neon transaction apply path.

## Possible approaches

1. Short-term: detect enum-value additions plus dependent same-plan statements, then reject the migration preview/apply with a clear precondition error telling the user to apply the enum label change first and regenerate the preview.
2. Longer-term: represent schema diff output as transaction groups, so enum value additions can commit before dependent statements while making the partial-apply semantics explicit in the UI/storage/apply flow.

The second option is more complete but needs careful design because splitting transactions changes the current all-or-nothing migration behavior.

## Testing ideas

- Add a unit test around migration validation/generation with `ALTER TYPE ... ADD VALUE` followed by a default or check expression using the new value.
- Add a regression test showing the dependent use is rejected clearly, or that transaction groups are generated/applied in the expected order if multi-transaction plans are implemented.

SSDWGG:
## Summary
- add an opt-in ts-pg-schema-diff guard that rejects plans which add a PostgreSQL enum value and use that new label later in the same transaction
- enable the guard for Dyad Neon migration previews, where apply runs the statement list in one transaction
- document the transactional enum caveat and cover the default-value regression path

## Why
PostgreSQL does not allow a newly added enum value to be used until the transaction that added it commits. Dyad currently applies Neon migration previews as one transaction, so plans like ALTER TYPE ADD VALUE followed by an ALTER TABLE SET DEFAULT using that value can look valid in preview and then fail at apply time. This turns that case into an early, clear precondition error that tells the user to apply the enum label change first and regenerate the preview.

Closes #3520

## Tests
- npm --prefix packages/ts-pg-schema-diff test
- npm --prefix packages/ts-pg-schema-diff run typecheck
- npm test -- src/ipc/utils/migration_utils.test.ts
- npm run fmt:check
- npm run ts

### DIFF
--- a/packages/ts-pg-schema-diff/README.md
+++ b/packages/ts-pg-schema-diff/README.md
@@ -46,7 +46,7 @@ type SchemaDiffResult = {
 ```
 
 Execute statements sequentially against `currentDatabaseUrl` to move it toward `desiredDatabaseUrl`.
-Do not wrap the whole result in one transaction: PostgreSQL rejects `CREATE INDEX CONCURRENTLY` and `DROP INDEX CONCURRENTLY` inside transaction blocks. If your executor requires a single transaction, call `generateSchemaDiff({ currentDatabaseUrl, desiredDatabaseUrl, noConcurrentIndexOperations: true })` and review the stronger locking behavior before applying.
+Do not wrap the whole result in one transaction: PostgreSQL rejects `CREATE INDEX CONCURRENTLY` and `DROP INDEX CONCURRENTLY` inside transaction blocks. If your executor requires a single transaction, call `generateSchemaDiff({ currentDatabaseUrl, desiredDatabaseUrl, noConcurrentIndexOperations: true, rejectEnumValueUsageInSameTransaction: true })` and review the stronger locking behavior before applying. The enum guard rejects plans that add a new enum value and use that value in a later statement, because PostgreSQL does not allow the value to be used until the transaction that added it commits.
 
 Connection options are passed to the underlying `pg` pool for both databases. The default pool size is `1` per database because introspection only needs one checked-out connection per side.
 
--- a/packages/ts-pg-schema-diff/src/generators/schema.ts
+++ b/packages/ts-pg-schema-diff/src/generators/schema.ts
@@ -578,8 +578,17 @@ function checkConstraintUsesAddedEnumLabel(
   constraint: CheckConstraint,
   addition: EnumValueAddition,
 ): string | null {
+  const enumColumnNames = new Set(
+    table.columns
+      .filter((column) => columnUsesType(column, addition.typeNames))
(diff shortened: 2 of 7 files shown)


SSDWGG:
Thanks for the review. I added a defensive null/undefined guard in findAddedLabelReference and reran the package tests/typecheck.

SSDWGG:

SSDWGG:
recheck

SSDWGG:
I signed the CLA in the requested format and posted `recheck`. The CLA bot comment now says all contributors have signed, but the GitHub check is still showing the old failed `CLAAssistant` result. Could a maintainer please rerun or refresh the CLA check when convenient?

SSDWGG:
Addressed the CHECK-constraint false positive identified by cubic.

The enum same-transaction guard now narrows CHECK constraint scanning to constraints whose `keyColumns` include one of the enum-typed columns. If column dependency metadata is absent, it keeps the previous conservative behavior. I also added coverage for both cases:

- a CHECK constraint on the enum column using the new label is still rejected
- an unrelated CHECK constraint on another column containing the same string literal is ignored

Local verification:

- `npm run typecheck` in `packages/ts-pg-schema-diff`
- `npm test` in `packages/ts-pg-schema-diff`
- `npx oxfmt --check packages/ts-pg-schema-diff/src/generators/schema.ts packages/ts-pg-schema-diff/test/unit.test.ts`

wwwillchen:
i don't think this is the right approach, we should have multiple transactions rather than outright rejecting

PR #3550, opened 2026-05-31
test: add validation test for prompt guide files

Kartalops:
Add test to validate prompt guide markdown files follow expected structure and format.


Kartalops:
Thank you for the detailed review! Both suggestions have been addressed:

1. `content.startsWith("## ")` → `content.trimStart().startsWith("## ")` — handles BOM, leading whitespace, newlines
2. Removed early-exit checks — files with orphan closing tags are now properly caught (expects 0 closing tags when no opening tags exist)

Fix has been force-pushed to the same branch.

wwwillchen:
we already have test coverage in  src/prompts/neon_prompt.test.ts

PR #3576, opened 2026-06-04
Automatically include jiti when installing nitro

wwwillchen:
Otherwise apps may end up transitively depending on jiti < v2 which causes issues on startup


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

PR #3592, opened 2026-06-05
fix(preview): rewrite Set-Cookie to SameSite=None in the proxy

azizmejri1:
The proxy now rewrites every forwarded Set-Cookie to
`SameSite=None; Secure; Partitioned` so cookies are sent inside the
preview iframe. This matters in packaged builds: the shell loads from a
file:// URL, a cross-site top-level ancestor that makes the request's
site-for-cookies cross-site, so default Lax/Strict cookies get dropped
and login breaks. Dev mode is unaffected since the shell is also
http://localhost.


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

azizmejri1:

### PR Confidence: 5/5
All three review threads were valid concerns addressed with code changes; tests, lint, and type-check pass.

### Unresolved Threads

_No unresolved threads_

### Resolved Threads
| Issue | Rationale | Link |
|-------|-----------|------|
| Malformed `Set-Cookie` from empty/non-string input or trailing/double semicolons (gemini-code-assist + dyad-assistant) | Added a non-string/empty guard and `filter(Boolean)` on split segments in `rewriteCookieForIframe`, so stray semicolons never produce `...; ; Secure` and empty input is returned unchanged | [View](https://github.com/dyad-sh/dyad/pull/3592#discussion_r3359844532) · [View](https://github.com/dyad-sh/dyad/pull/3592#discussion_r3359855254) |
| HTML-injection response path not covered by cookie tests (dyad-assistant) | Generalized the test upstream to serve `text/html` and added a test that exercises the injection branch (extensionless path + `<head>`), asserting cookies are still rewritten | [View](https://github.com/dyad-sh/dyad/pull/3592#discussion_r3359855251) |

<details>
<summary>Product Principle Suggestions</summary>

_No suggestions_ — all threads were unambiguous code-quality concerns. (Note: this skill installation has no `rules/product-principles.md`, so principle citations were unavailable; decisions used standard engineering judgment.)

</details>

---
🤖 Generated by Claude Code

wwwillchen:
@BugBot run

wwwillchen (commented):
@azizmejri1 - this is a great fix! thanks for debugging this.

i'm wondering if we should _not_ do partitioned. For example, if a user logs into the page inside the dyad preview iframe and then opens the preview in a new browser window, it would be nice for the user to stay logged in. IIUC, with partitioned, they would _not_ stay logged in because of the top-level origin (i.e. "file:" in the case of electron embed and "localhost" in the new browser window). If we did non-partitioned, then i believe they would stay logged in.

WDYT?

wwwillchen:
@BugBot run

wwwillchen (approved):
feel free to merge when ready

azizmejri1:
> @azizmejri1 - this is a great fix! thanks for debugging this.
> 
> i'm wondering if we should _not_ do partitioned. For example, if a user logs into the page inside the dyad preview iframe and then opens the preview in a new browser window, it would be nice for the user to stay logged in. IIUC, with partitioned, they would _not_ stay logged in because of the top-level origin (i.e. "file:" in the case of electron embed and "localhost" in the new browser window). If we did non-partitioned, then i believe they would stay logged in.
> 
> WDYT?

Hey @wwwillchen , since cookies are stored per browser even with non-partitioned, users would not stay logged in . However, after thinking about more partitioned attribute does not have a real benifit in our case since the preview iframe always runs inside the Dyad app, so there's only one top-level page and nothing to partition . I just dropped it and only kept SameSite=None; Secure.

wwwillchen:
@azizmejri1 - got it, OK lets just drop it then.

wwwillchen:
@BugBot run

PR #3616, opened 2026-06-08
Show release notes when restoring chat on startup

wwwillchen:
> [!NOTE]
> **Low Risk**
> UI-only refactor with localized settings behavior change when no release note exists; covered by new unit tests.
> 
> **Overview**
> **Release notes now run from the root layout** instead of the home page, so the “What’s new” dialog can appear when the app opens on chat or other routes—not only on home.
> 
> The logic lives in a new **`ReleaseNotesDialog`** component mounted next to the global toaster. It still checks once per session, skips the modal on first launch (but records the current version), and opens an iframe when `ipc.system.doesReleaseNoteExist` returns a URL. **Settings are only updated when notes are shown or on that first-run baseline**; if there is no release note for the version, `lastShownReleaseNotesVersion` is left unchanged (unlike the previous home-page flow, which always wrote the version before deciding whether to show the dialog).
> 
> Home page release-notes state, effects, and dialog markup were removed. **Vitest coverage** was added for upgrade, first-run, and missing-note cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14842adeef6e15904f48a529b38d1bb298c8fcde. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -22,6 +22,7 @@ import i18n from "@/i18n";
 import { LanguageSchema } from "@/lib/schemas";
 import { useShortcut } from "@/hooks/useShortcut";
 import { useIsMac } from "@/hooks/useChatModeToggle";
+import { ReleaseNotesDialog } from "@/components/ReleaseNotesDialog";
 
 export default function RootLayout({ children }: { children: ReactNode }) {
   const { refreshAppIframe } = useRunApp();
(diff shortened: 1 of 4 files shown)


wwwillchen:
@BugBot run

Issue #3552, opened 2026-06-01
dyad cannot connect to git repo suddenly even though it exists


ClickSendAI:
Error invoking remote method 'github:connect-existing-repo': Error: Failed to add remote: fatal: not a git repository (or any of the parent directories): .git

wwwillchen:
@ClickSendAI is the current app in Dyad a git repository? it looks like something got corrupted. could you try importing the app using the **Import** button on the home screen?

PR #3621, opened 2026-06-09
fix: clearer error when connecting to a non-git folder

Kartalops:
When the user picks a folder that is not a git repository and tries to connect it as an existing GitHub repo, the wrapped stderr `fatal: not a git repository (or any of the parent directories): .git` is forwarded verbatim, which is opaque (see #3552).

This detects the `not a git repository` case in `gitSetRemoteUrl` and surfaces an actionable message ("Initialize it with `git init` …") tagged as `DyadErrorKind.NotFound` instead of `Conflict`, matching the project's existing `DyadErrorKind` taxonomy. Other failure modes keep the original `Failed to add remote: …` text and Conflict kind.

### DIFF
--- a/src/ipc/utils/git_utils.ts
+++ b/src/ipc/utils/git_utils.ts
@@ -1022,10 +1022,15 @@ export async function gitSetRemoteUrl({
         }
       } else if (result.exitCode !== 0) {
         // Handle other errors
-        throw new DyadError(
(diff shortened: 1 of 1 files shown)


wwwillchen:
hi @Arvuno - thanks for the PR, could you respond with "I have read the CLA Document and I hereby sign the CLA" to sign-off on this PR? thanks

PR #3640, opened 2026-06-15
fix(agent-consent): show full SQL preview and restore "Show more"

RyanGroch:
The execute_sql consent preview was truncated to 100 characters before being handed to the consent banner. The banner decides whether to render its "Show more" button by checking if the preview overflows the collapsed box, so the 100-character cap kept the content short enough that it never overflowed. As a result "Show more" never appeared for SQL previews and longer queries were silently cut off with no way to see the rest.

Drop the character truncation and pass the full query. Replace the banner's JavaScript line measurement (getComputedStyle line-height, padding and box-sizing math) with a CSS line-clamp, so the collapsed preview clips cleanly on whole-line boundaries and "Show more" shows whenever the content actually overflows. Overflow detection is now a plain scrollHeight vs clientHeight check.

The banner is shared by all local-agent tool consents, but in practice only execute_sql and set_chat_summary produce previews long enough to overflow; every other tool renders a short one-line summary, so their appearance is unchanged.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/3640?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/components/chat/AgentConsentBanner.tsx
+++ b/src/components/chat/AgentConsentBanner.tsx
@@ -40,6 +40,8 @@ export function AgentConsentBanner({
   const [isInputExpanded, setIsInputExpanded] = React.useState(false);
   const [inputHasOverflow, setInputHasOverflow] = React.useState(false);
   const inputRef = React.useRef<HTMLDivElement | null>(null);
+  const isExpandedRef = React.useRef(isInputExpanded);
+  isExpandedRef.current = isInputExpanded;
 
   React.useEffect(() => {
     if (!inputPreview) {
@@ -50,9 +52,10 @@ export function AgentConsentBanner({
     const element = inputRef.current;
     if (!element) return;
 
-    // The collapsed preview is clipped to whole lines via CSS line-clamp, so
-    // overflow just means the full content is taller than the clamped box.
+    // Only measure while collapsed; the expanded box has different dimensions.
+    // Overflow means the full content is taller than the clamped preview.
     const compute = () => {
+      if (isExpandedRef.current) return;
       setInputHasOverflow(element.scrollHeight > element.clientHeight + 1);
     };
 
--- a/src/components/chat/AgentConsentBanner.tsx
+++ b/src/components/chat/AgentConsentBanner.tsx
@@ -40,9 +40,9 @@ export function AgentConsentBanner({
   const [isInputExpanded, setIsInputExpanded] = React.useState(false);
   const [inputHasOverflow, setInputHasOverflow] = React.useState(false);
   const inputRef = React.useRef<HTMLDivElement | null>(null);
-  const isExpandedRef = React.useRef(isInputExpanded);
-  isExpandedRef.current = isInputExpanded;
 
+  // Measure overflow only while collapsed; the expanded box has different
+  // dimensions. Re-runs on collapse so the flag refreshes once the box shrinks.
   React.useEffect(() => {
     if (!inputPreview) {
       setInputHasOverflow(false);
@@ -52,17 +52,15 @@ export function AgentConsentBanner({
     const element = inputRef.current;
     if (!element) return;
 
-    // Only measure while collapsed; the expanded box has different dimensions.
-    // Overflow means the full content is taller than the clamped preview.
     const compute = () => {
-      if (isExpandedRef.current) return;
+      if (isInputExpanded) return;
       setInputHasOverflow(element.scrollHeight > element.clientHeight + 1);
     };
 
     compute();
     window.addEventListener("resize", compute);
     return () => window.removeEventListener("resize", compute);
-  }, [inputPreview]);
+  }, [inputPreview, isInputExpanded]);
 
   return (
     <div className="border-b border-border bg-muted/50">
@@ -111,7 +109,7 @@ export function AgentConsentBanner({
                 className={`text-sm whitespace-pre-wrap ${
                   isInputExpanded
                     ? "max-h-[40vh] overflow-auto"
-                    : "line-clamp-4 overflow-hidden"
+                    : "line-clamp-6 overflow-hidden"
                 }`}
               >
                 {inputPreview}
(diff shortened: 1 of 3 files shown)


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

RyanGroch:
## ✅ Review feedback addressed

Pushed in `aa241b37`.

### Fixed
- **HIGH + cubic P2 (same root cause): resize while expanded could hide the collapse toggle.** Overflow detection compared `scrollHeight` to `clientHeight` against the element's current size, so a window resize while expanded could clear the overflow flag and remove the toggle, trapping the user in the expanded view. The check is now skipped while expanded, so the collapsed measurement stands.
- **MEDIUM: no upper bound on preview size.** Added a 2000 character cap on the SQL consent preview so a pathological query cannot push a multi-megabyte string through IPC and React state. The collapsed box still clips to a few lines and scrolls when expanded, so the cap is well above what is shown. Added unit tests for the full and capped cases.

### Intentional, keeping as is
- **MEDIUM: 4 vs 6 lines.** Keeping `line-clamp-4`. The MCP consent toast preview already uses a 4 line clamp (`McpConsentToast.tsx:55`), so 4 keeps the two consent surfaces consistent. The old value of 6 came from the previous pixel based clamp, not a deliberate target.

### CI note for a human
- `build (macos)` fails on a formatting check for `.claude/skills/pr-push/SKILL.md`. That file is identical to `main` on this branch, so the oxfmt drift is pre-existing and not introduced by this PR. `merge-reports` fails downstream of it (no e2e blob reports were produced), and `codex-review` is a known infra flake. None of these are caused by this change.

<details>
<summary>Files changed</summary>

- `src/components/chat/AgentConsentBanner.tsx`
- `src/pro/main/ipc/handlers/local_agent/tools/execute_sql.ts`
- `src/pro/main/ipc/handlers/local_agent/tools/execute_sql.spec.ts`

</details>

RyanGroch:
## ✅ Next round addressed

Pushed in `03e8b059`.

### Fixed
- **Stale overflow after collapse (cubic P2 + the resize-while-expanded MEDIUM).** Skipping the measurement while expanded left the overflow flag stale if the user expanded, resized so the content fit, then collapsed, keeping "Show more" visible until the next resize. The effect now also depends on `isInputExpanded`, so it re-measures on collapse. The earlier resize-while-expanded fix is preserved: `compute()` still returns early while expanded, so a resize in that state cannot clear the flag. Dropped the extra ref in favor of reading `isInputExpanded` directly.
- **Collapsed preview back to 6 lines.** Restored `line-clamp-6` to match the original behavior.

### CI
- `build (macos)` formatting failure is pre-existing drift in `.claude/skills/pr-push/SKILL.md` (identical to `main`), left as is per maintainer agreement. `merge-reports` cascades from it and `codex-review` is a known infra flake.

<details>
<summary>Files changed</summary>

- `src/components/chat/AgentConsentBanner.tsx`

</details>

wwwillchen:
@BugBot run

RyanGroch:
## ✅ Dropped the SQL preview cap

Pushed in `b294a882`. Addresses codex P1 (do not truncate SQL consent previews).

### Changed
- **Removed the 2000 character cap on the SQL consent preview.** It was defending against a near-impossible case: `args.query` comes from model tool output, so its size is already bounded, and Electron IPC handles that fine. Meanwhile the cap actively harmed the real case by hiding statements (including destructive ones) at the exact step where the user reviews the SQL before approving. `getConsentPreview` now returns the full query.
- The rendered size is still bounded where it matters: the banner clips to 6 lines collapsed and scrolls within `max-h-[40vh]` when expanded.
- Updated the unit test to assert a long query passes through untruncated.

### CI
- `build (macos)` formatting failure is pre-existing drift in `.claude/skills/pr-push/SKILL.md` (identical to `main`), left as is per maintainer agreement. `merge-reports` cascades from it; `codex-review` is a known infra flake.

<details>
<summary>Files changed</summary>

- `src/pro/main/ipc/handlers/local_agent/tools/execute_sql.ts`
- `src/pro/main/ipc/handlers/local_agent/tools/execute_sql.spec.ts`

</details>

wwwillchen:
@BugBot run

RyanGroch:
## ✅ Round addressed

Pushed in `1632ff08`.

### Fixed
- **codex P2: horizontal preview overflow.** A long unbroken token (large base64 or JSON literal, the 5000-char test case, etc.) did not wrap under `whitespace-pre-wrap`, so it overflowed horizontally without adding visual lines. The vertical overflow check stayed false, "Show more" never appeared, and the collapsed box clipped the tail. Added `break-words` so long tokens wrap. Wrapping keeps the SQL readable (better than a horizontal scrollbar for a consent prompt) and restores the existing vertical overflow detection, so no separate width measurement is needed.
- **MEDIUM: useLayoutEffect.** Moved the measurement from `useEffect` to `useLayoutEffect`. It reads geometry after the DOM update and before paint, which is the correct tool for a layout read driven by a class change. One-word change, no refactor.

### CI
- `build (macos)` formatting failure is pre-existing drift in `.claude/skills/pr-push/SKILL.md` (identical to `main`), left as is per maintainer agreement. `merge-reports` cascades from it; `codex-review` is a known infra flake.

<details>
<summary>Files changed</summary>

- `src/components/chat/AgentConsentBanner.tsx`

</details>

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen (approved):
thanks! i added a height animation so it looks a bit smoother. merging now.

Issue #3263, opened 2026-04-23
Local models (LM Studio) not loading


RWOverdijk:
**Describe the bug**

Try to select a local model from LM studio and the chat says "No loaded models found" with:

> Ensure LM Studio is running and models are loaded.

But LM Studio is running, and I can see the API call for models coming in, also:

```
2026-04-23 18:43:39 [DEBUG]
 Received request: GET to /api/v0/models
2026-04-23 18:43:39  [INFO]
 Returning {
  "data": [
    {
      "id": "qwen/qwen3.6-35b-a3b",
      "object": "model",
      "type": "vlm",
      "publisher": "qwen",
      "arch": "qwen35moe",
      "compatibility_type": "gguf",
      "quantization": "Q4_K_M",
      "state": "not-loaded",
      "max_context_length": 262144,
      "capabilities": [
        "tool_use"
      ]
    },
    {
      "id": "google/gemma-4-e4b",
      "object": "model",
      "type": "vlm",
      "publisher": "google",
      "arch": "gemma4",
      "compatibility_type": "gguf",
      "quantization": "Q8_0",
      "state": "not-loaded",
      "max_context_length": 131072,
      "capabilities": [
        "tool_use"
      ]
    },
    {
      "id": "google/gemma-4-26b-a4b",
      "object": "model",
      "type": "vlm",
      "publisher": "google",
      "arch": "gemma4",
      "compatibility_type": "gguf",
      "quantization": "Q4_K_M",
      "state": "not-loaded",
      "max_context_length": 262144,
      "capabilities": [
        "tool_use"
      ]
    },
    {
      "id": "text-embedding-nomic-embed-text-v1.5",
      "object": "model",
      "type": "embeddings",
      "publisher": "nomic-ai",
      "arch": "nomic-bert",
      "compatibility_type": "gguf",
      "quantization": "Q4_K_M",
      "state": "not-loaded",
      "max_context_length": 2048
    }
  ],
  "object": "list"
}
```

So unsure why dyad says no models are loaded when it clearly gets a result containing models.

**To Reproduce**

Download lm studio and dyad, download some models in lm studio. Try to start a chat with a local model in dyad.

**Expected behavior**

For local models to be discoverable and usable

**Desktop (please complete the following information):**

- OS: MacOS 
- Version: 0.43.0

**Additional context**

Models work fine in OpenCode and in LM Studio.

RWOverdijk:
[...]

PR #3669, opened 2026-06-22
fix(lmstudio): list vlm and other generative models, not just llm

### LINKED ISSUE #3668: LM Studio models return empty list due to strict `"llm"` filter
Related issue: [https://github.com/dyad-sh/dyad/issues/3263](https://github.com/dyad-sh/dyad/issues/3263)

In this file:

[https://github.com/dyad-sh/dyad/blob/93fc2775e759d8497104221d2385eb5c6d69b5ac/src/ipc/handlers/local_model_lmstudio_handler.ts#L36](https://github.com/dyad-sh/dyad/blob/93fc2775e759d8497104221d2385eb5c6d69b5ac/src/ipc/handlers/local_model_lmstudio_handler.ts#L36)

the code filters models using:

```ts id="gq9n8p"
filter((model) => model.type === "llm")
```

However, LM Studio’s `/api/v0/models` endpoint returns models that are not labeled `"llm"`, even when they are usable chat models. I assume `"vlm"` is in reference to the model also being vision capable (https://lmstudio.ai/docs/typescript/llm-prediction/image-input). 

### Example response from LM Studio

```json id="k2v8sn"
{
  "data": [
    {
      "id": "google/gemma-4-e4b",
      "object": "model",
      "type": "vlm",
      "publisher": "google",
      "arch": "gemma4",
      "compatibility_type": "mlx",
      "quantization": "4bit",
      "state": "loaded",
      "max_context_length": 131072,
      "loaded_context_length": 8192,
      "capabilities": [
        "tool_use"
      ]
    },
    {
      "id": "text-embedding-nomic-embed-text-v1.5",
      "object": "model",
      "type": "embeddings",
      "publisher": "nomic-ai",
      "arch": "nomic-bert",
      "compatibility_type": "gguf",
      "quantization": "Q4_K_M",
      "state": "not-loaded",
      "max_context_length": 2048
    }
  ],
  "object": "list"
}
```

### Observed behavior

Because none of the returned models have `type: "llm"`, the filter removes everything, resulting in:

```text id="v7p2lx"
Successfully fetched 0 models from LM Studio
```

This can also be confirmed in logs (macOS):

```
~/Library/Logs/Dyad/main.log

[2026-06-20 21:56:05.866] [info] (lmstudio_handler) Successfully fetched 0 models from LM Studio
```

---

### Suggested fix

A minimal fix is to include `"vlm"`:

```ts id="z1q3ta"
filter((model) => model.type === "llm" || model.type === "vlm")
```

Alternatively, a more robust approach may be to exclude only embeddings:

```ts id="m8x4ld"
filter((model) => model.type !== "embeddings")
```

This should allow LM Studio chat-capable models (including multimodal ones) to appear correctly.

---

serhiizghama:
## Problem

Fixes #3668.

LM Studio's `/api/v0/models` endpoint reports model `type` as `"llm"`, `"vlm"` (multimodal/vision-capable chat models), or `"embeddings"`. The handler filtered with:

```ts
.filter((model) => model.type === "llm")
```

So any chat model tagged `"vlm"` was dropped. A user whose loaded models are all multimodal ends up with an empty list and this log line:

```
(lmstudio_handler) Successfully fetched 0 models from LM Studio
```

This is the same root cause as the previously reported #3263.

## Solution

Exclude embedding model types instead of allow-listing `"llm"`:

```ts
const NON_CHAT_LM_STUDIO_MODEL_TYPES = new Set(["embeddings", "embedding"]);
...
.filter((model) => !NON_CHAT_LM_STUDIO_MODEL_TYPES.has(model.type))
```

Both `"embeddings"` (current API value) and the legacy singular `"embedding"` (already declared in the `LMStudioModel` type) are excluded. Allow-listing `"llm"` was the bug; blocking only the non-chat (embedding) types keeps `vlm` and any future chat-capable type usable without another code change, while embedding models stay hidden. The `LMStudioModel` `type` union is also updated to match the values the endpoint actually returns.

## Testing

Added `local_model_lmstudio_handler.test.ts` (mirrors the existing Ollama handler test, stubs `fetch`):

- a mixed list of `vlm` + `llm` + `embeddings` returns only the two generative models
- the legacy singular `embedding` type is also dropped
- a non-ok response still throws an external `DyadError`

```
✓ src/ipc/handlers/local_model_lmstudio_handler.test.ts (3 tests)
```

Reverting the filter to `type === "llm"` makes the vlm case fail, confirming the test guards the fix. `oxlint` and `tsgo` type-check are clean on the changed files.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/3669?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/ipc/handlers/local_model_lmstudio_handler.ts
+++ b/src/ipc/handlers/local_model_lmstudio_handler.ts
@@ -7,8 +7,15 @@ import { DyadError, DyadErrorKind } from "@/errors/dyad_error";
 
 const logger = log.scope("lmstudio_handler");
 
+// LM Studio's /api/v0/models endpoint returns both generative models and
+// embedding models. Only generative models can be used for chat, so the
+// embedding types are excluded. Excluding non-chat types (rather than
+// allow-listing "llm") keeps multimodal/vision models such as "vlm" usable
+// and lets future chat-capable types surface without another code change.
+const NON_CHAT_LM_STUDIO_MODEL_TYPES = new Set(["embeddings", "embedding"]);
+
 export interface LMStudioModel {
-  type: "llm" | "embedding" | string;
+  type: "llm" | "vlm" | "embeddings" | string;
   id: string;
   object: string;
   publisher: string;
@@ -33,8 +40,8 @@ export async function fetchLMStudioModels(): Promise<{ models: LocalModel[] }> {
   const modelsJson = await modelsResponse.json();
   const downloadedModels = modelsJson.data as LMStudioModel[];
   const models: LocalModel[] = downloadedModels
-    .filter((model: any) => model.type === "llm")
-    .map((model: any) => ({
+    .filter((model) => !NON_CHAT_LM_STUDIO_MODEL_TYPES.has(model.type))
+    .map((model) => ({
       modelName: model.id,
       displayName: model.id,
       provider: "lmstudio",
--- a/src/ipc/handlers/local_model_lmstudio_handler.test.ts
+++ b/src/ipc/handlers/local_model_lmstudio_handler.test.ts
@@ -1,0 +1,99 @@
+import { DyadError, DyadErrorKind } from "@/errors/dyad_error";
+import { fetchLMStudioModels } from "@/ipc/handlers/local_model_lmstudio_handler";
+import { afterEach, describe, it, expect, vi } from "vitest";
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+  vi.restoreAllMocks();
+});
+
+function stubFetchWith(data: unknown[]) {
+  vi.stubGlobal(
+    "fetch",
+    vi.fn().mockResolvedValue({
+      ok: true,
+      json: vi.fn().mockResolvedValue({ data, object: "list" }),
+    }),
+  );
+}
+
+describe("fetchLMStudioModels", () => {
+  it("includes generative models (llm and vlm) and drops embedding models", async () => {
+    // Example response shape taken from LM Studio's /api/v0/models endpoint.
+    stubFetchWith([
+      {
+        id: "google/gemma-4-e4b",
+        object: "model",
+        type: "vlm",
+        publisher: "google",
+        state: "loaded",
+      },
+      {
+        id: "qwen/qwen3-8b",
+        object: "model",
+        type: "llm",
+        publisher: "qwen",
+        state: "loaded",
+      },
+      {
+        id: "text-embedding-nomic-embed-text-v1.5",
+        object: "model",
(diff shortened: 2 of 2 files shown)


wwwillchen:
Thanks @serhiizghama - changes looks good to me!

Can you reply with a comment that says:

```
```

serhiizghama:

src/ipc/handlers/local_model_lmstudio_handler.ts
```diff
@@ -7,8 +7,15 @@ import { DyadError, DyadErrorKind } from "@/errors/dyad_error";
...
+// LM Studio's /api/v0/models endpoint returns both generative models and
+// embedding models. Only generative models can be used for chat, so the
+// embedding types are excluded. Excluding non-chat types (rather than
+// allow-listing "llm") keeps multimodal/vision models such as "vlm" usable
+// and lets future chat-capable types surface without another code change.
+const NON_CHAT_LM_STUDIO_MODEL_TYPES = new Set(["embeddings", "embedding"]);
+
 export interface LMStudioModel {
-  type: "llm" | "embedding" | string;
+  type: "llm" | "vlm" | "embeddings" | string;
```
wwwillchen, line 18: i looked at http://lmstudio.ai/docs/developer/rest/list?utm_source=chatgpt.com and it doesn't show "vlm" as a model type - i don't think this PR is needed?

serhiizghama: That /rest/list page just has an abbreviated example — the full endpoints reference at https://lmstudio.ai/docs/developer/rest/endpoints does list it, with a `qwen2-vl-7b-instruct` entry showing `"type": "vlm"` right next to `llm` and `embeddings`. And it's not just docs: both #3668 and the earlier #3263 include real LM Studio server logs where every loaded model comes back as `vlm` (e.g. `google/gemma-4-e4b`, `qwen/qwen3.6-35b-a3b`), which is exactly why those users hit `fetched 0 models`. So the `type === "llm"` allow-list does drop real chat models today — that's what this narrows to just excluding embeddings.

wwwillchen: got it, thanks for clarifying! merging

wwwillchen (approved):
lgtm

PR #3769, opened 2026-07-03
Keep cancelled messages in LLM context

### LINKED ISSUE #3756: Cancelled messages should be visible to LLM as context
I think we originally discussed hiding cancelled messages from the LLM for future turns, but I think it actually creates kind of weird edge cases. For example, sometimes Dyad LLM response will hang so then the user will cancel. If you try to say "keep going", it won't know what was done in the previous (partially done) turn and get confused.

Example:

Image

@azizmejri1 could you look into this?

azizmejri1:
closes #3756 
Previously the cancelled assistant message and its triggering user message were filtered out of the history via filterCancelledMessagePairs before building context for the next request. Stop filtering them so the model now sees the interrupted turn (the partial response still ends with the "[Response cancelled by user]" marker) and the prompt that triggered it.

Removes the now-dead filterCancelledMessagePairs helper and its unit tests, and flips the cancelled-message e2e assertions to expect the content to be present in the dumped LLM context.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/3769?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/e2e-tests/cancelled_message.spec.ts
+++ b/e2e-tests/cancelled_message.spec.ts
@@ -2,7 +2,7 @@ import { test } from "./helpers/test_helper";
 import { expect } from "@playwright/test";
 import fs from "fs";
 
-test("cancelled message shows cancelled indicator and is excluded from context", async ({
+test("cancelled message shows cancelled indicator and is kept in context", async ({
   po,
 }) => {
   await po.setUp();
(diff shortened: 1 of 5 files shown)


wwwillchen:
@BugBot run

wwwillchen:
thanks!

PR #3781, opened 2026-07-06
Fix invisible code view toolbar icons on hover in dark mode

azizmejri1:
Add dark:hover:bg-gray-700 to the refresh and fullscreen buttons so the hover background is dark instead of light gray, keeping the icons visible.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/3781?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/components/preview_panel/CodeView.tsx
+++ b/src/components/preview_panel/CodeView.tsx
@@ -81,7 +81,7 @@ export const CodeView = ({ loading, app }: CodeViewProps) => {
               render={
                 <button
                   onClick={() => refreshApp()}
-                  className="p-1 rounded hover:bg-gray-200 disabled:opacity-50 disabled:cursor-not-allowed"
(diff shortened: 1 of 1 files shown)


wwwillchen:
@BugBot run

wwwillchen (approved):
thanks @azizmejri1! lgtm.

i also noticed in fullscreen mode - the buttons overlap with the mac chrome and the exit fullscreen mode button doesn't work. 
do you mind fixing this in a follow-up PR? I'm also OK if you want to just remove the fullscreen mode feature too (personally I don't use it and not sure how much it's really used).

Image

PR #3790, opened 2026-07-06
 fix: don't clear chat input when Enter-key submit is rejected

wwwillchen:
## Summary
- Fixes the onboarding bug where a prompt submitted with the Enter key (with no AI provider configured) was silently wiped: the setup dialog claimed "Your prompt is saved", but after adding an API key nothing auto-submitted, and the still-armed pending-prompt flag auto-submitted the user's next keystroke (e.g. "a") instead of the original prompt.
- Root cause was not a race: `LexicalChatInput`'s Enter path cleared the editor unconditionally, even when the submit was rejected and queued — while the send-button path did not. Removing the clear makes Enter identical to the button; the editor still clears via the existing `value === ""` effect since both callers clear the value on successful submit. This also stops the chat page from wiping input when queueing a message mid-stream fails.
- Adds an E2E regression test that submits the initial prompt with Enter (the previously untested path) and verifies it survives provider setup and auto-submits.

---

> [!NOTE]
> **Low Risk**
> Scoped UX fix to chat submit/clear behavior with documentation and E2E coverage; no auth or data-layer changes.
> 
> **Overview**
> Fixes onboarding where **Enter** could wipe the first home prompt even though the AI setup dialog said the prompt was saved—after adding a key, auto-resume had nothing to send and the next keystroke could submit instead of the original text.
> 
> **`LexicalChatInput`** no longer clears the editor inside its Enter submit wrapper; **`EnterKeyPlugin`** calls the parent **`onSubmit`** only. The editor still clears when the caller sets the controlled value to **`""`** (existing external-clear effect), matching the send-button path.
> 
> **`home.tsx`** clears **`inputValue`** immediately after **`streamMessage`** is dispatched (before later awaits), so a successful send clears right away while rejected submits (no provider / pending first prompt) leave the text in the box.
> 
> Adds an E2E test that submits with **`Enter`**, asserts the prompt stays visible during Google API setup, then verifies auto-resume after save. Docs note caller-owned clearing and that Enter vs send button are separate code paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b8bf560a66510facf5be1c6c69e7f88dc9c382cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/3790?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

wwwillchen:
@BugBot run

PR #3816, opened 2026-07-07
fix: flash when dialogs close

RyanGroch:
Closing any Dialog or AlertDialog briefly flashed the dark backdrop at the end of the fade-out. The exit animation runs with the default animation-fill-mode (none), so the overlay snaps back to full opacity for the frames between the animation ending and the element unmounting. Holding the animation's final frame with fill-mode-forwards removes the flash.

An easy way to observe the flash is by opening and closing the help dialog (question mark icon in the bottom left). Overall this is low-priority because it's just a small visual blip, but even so I think it's worth fixing.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/3816?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/components/ui/alert-dialog.tsx
+++ b/src/components/ui/alert-dialog.tsx
@@ -28,7 +28,7 @@ function AlertDialogOverlay({
     <AlertDialogPrimitive.Backdrop
       data-slot="alert-dialog-overlay"
       className={cn(
-        "data-open:animate-in data-closed:animate-out data-closed:fade-out-0 data-open:fade-in-0 fixed inset-0 z-50 bg-black/50",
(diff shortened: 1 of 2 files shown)


wwwillchen:
@BugBot run

wwwillchen (approved):
thanks!

PR #3909, opened 2026-07-12
Report Windows OOM crashes with a named crash reason

RyanGroch:
On Windows, when a Chromium-based process dies because a memory allocation failed, it raises a custom exception code: 0xE0000008 (kOomExceptionCode in Chromium's partition_alloc/oom.h). This covers native allocation failures and also V8 heap exhaustion in the main process, which Electron's OOM handler routes into the same termination path. Other platforms do not produce this code.

Our minidump parser didn't recognize the code, so these crashes showed up in the app:crash_detected telemetry event with an empty crash_reason and only a raw exception_code. Add the code to the parser's Windows table so they now report crash_reason: "OUT_OF_MEMORY".

This is one part of a broader improvement to OOM crash detection: it makes Windows OOM crashes self-identifying with no new machinery. A follow-up PR will add an explicit OOM classification (OOM or not, plus which evidence decided it) that combines this exception code with heap-usage data recorded before the crash, extending coverage to macOS and Linux.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/3909?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/utils/minidump_summary.test.ts
+++ b/src/utils/minidump_summary.test.ts
@@ -208,6 +208,20 @@ describe("parseMinidumpBuffer", () => {
     expect(s!.faultingModule).toBe("app.exe");
   });
 
+  it("decodes Chromium's Windows OOM exception code", () => {
+    // 0xE0000008 is kOomExceptionCode, raised only when a process is
+    // terminated because an allocation failed.
+    const dump = buildMinidump({
+      modules: [{ base: 0x400000n, size: 0x1000, name: "C:\\app\\app.exe" }],
+      exceptionCode: 0xe0000008,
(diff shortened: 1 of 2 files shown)


wwwillchen:
@BugBot run

wwwillchen (approved):
thanks!

PR #3956, opened 2026-07-16
Fix Supabase OAuth token propagation race

wwwillchen:
## Summary

Newly issued Supabase OAuth tokens can briefly return 401 from the Management API before token propagation completes, causing Dyad to miss organization-scoped credential setup. This gives the initial organization lookup a short retry window so the connection completes without requiring the user to reopen the deep link.

- Retry only 401 responses, using four bounded backoff delays totaling 3.75 seconds; existing 429 handling remains unchanged.
- Preserve immediate failure behavior for 403 and all other non-retryable responses. A genuinely invalid token may now take up to 3.75 seconds before the existing legacy fallback runs.
- Add unit coverage for transient-401 recovery and the non-401 boundary.
- Document the macOS custom-protocol ownership check that prevents false results while bisecting deep-link behavior.

---

> [!NOTE]
> **Low Risk**
> Small, bounded retry on one API helper plus docs and tests; invalid tokens may fail up to ~3.75s later before existing fallback.
> 
> **Overview**
> Fixes a race where **fresh Supabase OAuth tokens** can get **401** from the Management API during org lookup right after connect, which blocked organization-scoped credential setup.
> 
> **`listSupabaseOrganizations`** now retries **401 only** with four backoff delays (250ms–2s, ~3.75s total) before failing; **403** and other errors still fail immediately. Existing **429** behavior via `fetchWithRetry` is unchanged.
> 
> Adds **Vitest** coverage for transient-401 recovery vs non-retryable 403. **`rules/electron-ipc.md`** adds a note to verify **macOS `dyad://` protocol ownership** when bisecting deep-link flows so the wrong build does not fake good/bad results.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2be7689218351faebd2b093325841d1bb0ca7753. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/rules/electron-ipc.md
+++ b/rules/electron-ipc.md
@@ -71,6 +71,8 @@ writeSettings({
 
 **Electron readiness:** `readSettings()` and `writeSettings()` may decrypt/encrypt secrets through Electron `safeStorage`, which throws `safeStorage cannot be used before app is ready` before `app.whenReady()`. Queue pre-ready entry points like deep links (`open-url`, `second-instance`) until the app/window is ready before calling OAuth/settings handlers.
 
+**Custom-protocol debugging:** Before using `git bisect` on a `dyad://` flow, quit every dev and packaged Dyad instance and verify which build owns the protocol registration. macOS may route the link to a different running/registered build, producing a convincing but false good/bad result.
+
 ## Handler expectations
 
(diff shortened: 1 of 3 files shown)


wwwillchen:
@BugBot run

PR #4003, opened 2026-07-21
Replace VersionPane preview orchestration with explicit state machine

wwwillchen:
## Summary

Version History preview (checkout/return/restore/recovery) is now driven by a pure transition function plus a serial command controller in `src/version_preview/`, hosted above the pane in a per-app registry bound via `useSyncExternalStore`.

Fixes close-during-checkout, failed-return retry, app-switch abandonment, and diff-clearing recovery races. Design record: `plans/version-preview-state-machine.md`.


wwwillchen:
@BugBot run

PR #4087, opened 2026-07-24
Add delete and open-in-editor actions to the Tests panel

azizmejri1:
Each spec row in the Tests panel now carries two new actions alongside Run:

- Open in code editor (code icon), which switches the preview panel to the code view with the spec selected. Individual test rows get the same action and jump to that test's own line.
- Delete (trash icon), behind a confirmation that names the file and the number of tests going with it. It's disabled while a run is in flight.

Screenshot from 2026-07-24 21-10-17

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4087?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


azizmejri1:

### PR Confidence: 5/5
All three review threads were legitimate P2 issues, each addressed with focused code changes plus test coverage; type-check and the affected unit tests pass.

### Unresolved Threads
_No unresolved threads_

### Resolved Threads
| Issue | Rationale | Link |
|-------|-----------|------|
| Delete could slip through the confirmation window if a run started after the dialog opened | Guarded `confirmDelete` to bail while `isRunning`, and added an effect that closes the dialog when a run starts — the backend `withLock` already serialized deletes, so this closes the UI-level gap. Per **Principle #4: Transparent Over Magical**, the stated no-delete-during-run guarantee now holds through the confirmation window. | [View](https://github.com/dyad-sh/dyad/pull/4087#discussion_r3647940961) |
| Changes/diff UI stayed stale after deleting a tracked spec | `useDeleteAppTest` now also invalidates `uncommittedFiles.byApp`, so the staged deletion appears in the Changes diff immediately. Per **Principle #4: Transparent Over Magical**, the staged change is surfaced right away. | [View](https://github.com/dyad-sh/dyad/pull/4087#discussion_r3647940967) |
| Dialog/toast promised recoverability even for untracked files that are gone for good | Handler now returns `staged` (whether `gitRemove` succeeded); the toast reports "was untracked, not recoverable" when false, and the dialog copy is conditional ("If it's tracked in git…"). Added handler-test assertions for both branches. Per **Principle #4: Transparent Over Magical**, the UI no longer over-promises a recovery path it can't keep. | [View](https://github.com/dyad-sh/dyad/pull/4087#discussion_r3647940978) |

---
🤖 Generated by Claude Code

wwwillchen (approved):
overall looks good. couple of minor comments:

- for the open in code view, can we use the code icon? the current icon is mostly used for opening a new window (esp. browser).

Image

- it feels kind of weird to leave the deleted test files be staged. can we commit the deleted files? let's explicitly commit the deleted file (and not other files) so then users don't get the blue banner with "You have X uncommitted changes". i'm guessing you might have did this in case users want to batch delete multiple test files (and just have one commit), but we could do a checkbox (probably later) for multi-selection delete.

azizmejri1:

### PR Confidence: 4/5
All four review threads were real defects with contained fixes, each now covered by a regression test and a green unit suite (420 files, 4263 tests); the remaining uncertainty is that the delete flow hasn't been exercised end-to-end against a real repo.

### Unresolved Threads
_No unresolved threads_

### Resolved Threads
| Issue | Rationale | Link |
|-------|-----------|------|
| A concurrent save can be deleted after confirmation (`tests_handlers.ts`) | Dropped the `unlink`-then-`git rm -f` sequence: `git rm` now removes the file and stages the deletion in one operation, so a write landing in between can't be force-deleted. An untracked file (git removed nothing) still falls back to an `unlink`, and the handler pre-checks existence so a missing spec is still `NotFound`. Regression test asserts a file recreated during the git step is left alone. Per **Principle #4: Transparent Over Magical**, a destructive action must not extend past what the user confirmed. | [View](https://github.com/dyad-sh/dyad/pull/4087#discussion_r3667867142) |
| Bind delete confirmation to the originating app (`TestsPanel.tsx`) | `pendingDelete` now stores `{ appId, file }`, `confirmDelete` deletes against the originating app and bails if the selection moved, and an effect clears the pending confirmation on app switch. Test covers the dialog closing on app change without issuing a delete. | [View](https://github.com/dyad-sh/dyad/pull/4087#discussion_r3670068891) |
| Open-in-editor can remain stuck in a staged diff (`TestsPanel.tsx`) | `openInEditor` clears `stagedDiffFileAtom` before selecting the file, since `CodeView` renders a staged diff in preference to the selected file. Test asserts the atom is cleared. | [View](https://github.com/dyad-sh/dyad/pull/4087#discussion_r3670068894) |
| Don't report every commit failure as an untracked deletion (`useDeleteAppTest.ts`) | `removeFileAndCommit` now returns `{ commitHash, uncommittedReason }` (`"untracked"` \| `"commit-failed"` \| `null`), plumbed through the `tests:delete` contract. The toast says "was untracked, not recoverable" only for a genuinely untracked file, and "not committed — see pending changes" when the deletion is staged. Per **Principle #4: Transparent Over Magical**, recovery guidance after a destructive action has to match what actually happened. | [View](https://github.com/dyad-sh/dyad/pull/4087#discussion_r3670068895) |

---
🤖 Generated by Claude Code

PR #4127, opened 2026-07-28
feat: deploy Dyad apps to a self-hosted server with Coolify

RyanGroch:
*This description was written by Claude but is human-reviewed.*

**This PR is not meant to be merged.** It is a proof of concept, built to find
out whether a single API integration can replace a hand-rolled deployment
pipeline, and what the experience feels like end to end. It works against real
servers, but see *Known gaps* before treating any part as finished.

Adds a second deployment target alongside Vercel: a server the user owns,
running [Coolify](https://coolify.io). Coolify handles the proxy, TLS, build and
process supervision; Dyad drives its API.

Beyond hosting the app, Dyad can provision a Postgres database on that same
server and keep its schema in step with the app's development database. That
database is never exposed to the internet — Dyad reaches it through an SSH
tunnel to its container, which also supplies the transport encryption Coolify's
Postgres does not have.

To make an app portable enough to run against a self-hosted database, this also
adds a **Portable Postgres** option: the same Neon project during development,
but the model generates standard `pg` code instead of Neon-specific code, so the
app runs unchanged against any Postgres.

Finally, Dyad can install Coolify on a bare server over SSH, so setting one up
does not require a terminal.

## Benefits

- **It already deploys a full application, database included.** Not just static
hosting: an app, a Postgres database on the user's own server, and schema
migrations that keep the two in step.
- **The user never touches a terminal.** Dyad generates the SSH key, installs
Coolify, provisions the database, opens the tunnel and runs the migration.
- **A user-provided Postgres would be a small addition.** The production
database is already resolved through one path; accepting a connection string
is a field and a branch, and needs no tunnel because such a database is
already reachable.
- **Cloud databases still work.** An app on Neon or Supabase can deploy here and
keep using it — Coolify hosts the app only.

## Limitations

These are inherent to the approach rather than things left unfinished.

- **It needs a real server.** Coolify wants roughly 2 vCPU, 2 GB RAM and 30 GB
of storage before the user's app is built on the same machine. Smaller boxes
get their builds killed.
- **The user has to use Coolify's interface.** Not difficult, but it is a second
product with its own concepts — servers, projects, environments — and some
learning curve.
- **A hosted git repository is required.** Coolify clones over SSH, so the code
must live somewhere SSH-reachable. It can be private. A local-only repository
is not enough. Coolify supports GitHub, GitLab and Gitea including self-hosted
instances; this PR handles GitHub only.
- **The server must be reachable as root.** Both the install and the tunnel
assume it.
- **Everything is slower than a managed platform.** Coolify builds in Docker on
the user's server, so installs and deploys take minutes rather than seconds.

## Trying it

These instructions take the simplest path: **deploying to the same server that
runs Coolify, using a Postgres database that Coolify provisions there.** Two
things worth knowing that we will not be exercising here — the branch also
supports deploying to a **different** server from the one Coolify runs on, and
apps can keep using a **Neon or Supabase cloud database** instead of a
self-hosted one. Only self-hosted Supabase and Neon are out of scope.

_(Note from Ryan: I can make a video demo of this if that would be helpful)_

**1. Get a server.** Any Ubuntu/Debian VPS with at least 2 vCPU and 2 GB RAM.

**2. Give Dyad SSH access.** Open an app → **Publish** → **Deployment** →
**Your own server**. Click **Generate SSH key**, copy it, and paste it into your
provider's SSH-key field when creating the server (or append it to
`~/.ssh/authorized_keys` on an existing one).

**3. Install Coolify** — either way works:
- *From Dyad:* enter your email and the server's address, then click **Install
Coolify on this server**. Progress streams in the panel; when it finishes it
shows the dashboard URL and a generated password.
- *Yourself:* `curl -fsSL https://cdn.coollabs.io/coolify/install.sh | sudo bash`,
then create the admin account at `http://YOUR_IP:8000`.

**4. Finish Coolify's own setup.** Sign in and work through its onboarding. It
asks whether you want to deploy to the server Coolify is running on — say yes,
which is what these instructions assume.

**5. Create an API token.** Go to **Keys & Tokens → API tokens**. API access is
off by default, so you will be prompted to enable it in settings first; do that,
then come back to this page and create the token. It needs **four** scopes —
`read`, `read:sensitive`, `write`, `deploy`. Scopes are fixed at creation, so a
token missing one has to be replaced.

**6. Connect Dyad to Coolify.** Back in **Publish → Deployment → Your own
server**, paste the Coolify URL (already filled in if Dyad did the install) and
the token. Then pick your server and project from the dropdowns, click **Test
SSH**, and **Save**. The SSH fields fill themselves in from the server you pick.

**7. Set up the database.** On the app's details page, the database card has two
tabs: **Neon** and **Portable Postgres**. Choose **Portable Postgres** and click
through the sign-in. Pick this *before* building data features — it changes the
code the model writes, and switching later does not rewrite existing files.

**8. Build something that stores data**, and check the generated code uses `pg`
rather than `@neondatabase/serverless`.

**9. Connect GitHub** (a private repo is fine) and push. Coolify clones over
SSH; Dyad registers the deploy key for you.

**10. Deploy.** In **Publish → Deployment → Your own server**, click **Deploy**.
Expect: SSH check → database provisioned → schema migrated through the tunnel →
application created → build → a live URL. Open it and use the feature you built;
it should read and write against the database on your own server.

### A. SSH foundation — ~770 lines
- `src/ipc/utils/ssh_utils.ts` (344) — locates the OpenSSH binaries, manages a
dedicated key, classifies errors into actionable categories, runs commands
buffered or streaming
- `src/ipc/utils/ssh_tunnel.ts` (223) — resolves a container's address on the
server's Docker network and forwards a local port to it
- `ssh_utils.test.ts` (88), `ssh_tunnel.test.ts` (119)

### B. Coolify API client — ~820 lines
- `src/ipc/utils/coolify_client.ts` (364) — typed client for the endpoints used,
with error mapping for bad tokens, missing scopes and unreachable instances
- `src/ipc/utils/coolify_install.ts` (114) — credential generation and the
install command, with the shell quoting locked down
- `coolify_client.test.ts` (246), `coolify_install.test.ts` (94)

### C. Deployment orchestration — ~1,250 lines
- `src/ipc/handlers/coolify_handlers.ts` (920) — the deploy pipeline and the
install runner
- `src/ipc/types/coolify.ts` (261) — IPC contracts and events
- `src/shared/coolify_provisioning.ts` (17) + test (44) — decides whether Coolify
should host a database, shared so the UI and the deploy cannot disagree
- `src/ipc/handlers/github_handlers.ts` — exposes the API base for deploy-key
registration

### D. Deployment UI — ~890 lines
- `src/components/CoolifyConnector.tsx` (707), `src/hooks/useCoolifyDeploy.ts`
(185)
- `src/components/preview_panel/PublishPanel.tsx` — Vercel and Coolify merged
into one **Deployment** card with a tab each

### E. Portable Postgres — ~800 lines
- `src/prompts/portable_postgres_prompt.ts` (125) + test (61) — the codegen rules
- `src/postgres_admin/postgres_context.ts` (128) + test (129) — socket-based SQL
execution and transactional migration
- `src/postgres_admin/portable_postgres_prompt_context.ts` (43)
- `src/components/PostgresConnector.tsx` (264)
- `src/ipc/handlers/chat_stream_handlers.ts` — selects the prompt
- `src/pages/app-details.tsx` — the Neon / Portable Postgres tabs
- `src/components/NeonConnector.tsx` — an `embedded` prop so it can sit in a tab
- `src/hooks/useNeon.ts` — only query a Neon project once the user is signed in
- `ConfigurePanel.tsx`, `i18n/locales/en/home.json` — name the provider Postgres
when that is what was chosen

### F. Schema, settings and registration — ~200 lines
- `src/db/schema.ts`, `drizzle/0042_*`, `drizzle/0043_*`
- `src/main/settings.ts`, `src/lib/schemas.ts` (+ settings test) — encrypted
token and admin password
- `src/ipc/ipc_host.ts`, `preload/channels.ts`, `types/index.ts`,
`lib/queryKeys.ts`, `types/app.ts`

## Splitting this up

Each of these stands alone as a shippable feature, in this order:

**PR 1 — Portable Postgres** (component E, plus its column in F)
A third database option that generates standard Postgres code. Useful with no
deployment involved: it is what lets an app move between providers. Independent
of everything else here.

**PR 2 — Deploy to Coolify** (components B, D, the deployment half of C, rest of F)
Deploy an app to a server running Coolify. No SSH, no tunnel, no provisioning.
Complete for apps with no database, and for apps keeping a cloud database. The
SSH preflight currently in the deploy belongs in PR 3.

**PR 3 — Coolify-hosted database** (component A, plus the database half of C)
Provision Postgres on the user's server and migrate its schema through an SSH
tunnel. Needs PR 1 for an app that can use it and PR 2 for something to deploy.
This is where the subtle work is and deserves the most careful review.

**PR 4 — Install Coolify without a terminal** (`coolify_install.ts`,
`runRemoteStreaming`, the install UI)
Purely additive: everything else works with a Coolify the user installed
themselves.

If it stays as one PR, review it in that order.

## Known gaps

- **Creating the API token is manual.** Coolify has no API for minting the first
one.
- **The app's port is hardcoded to 3000.** Fine for Next.js; wrong for anything
else.
- **No custom domain field.** Apps get a generated `sslip.io` address over plain
HTTP. A domain can be set in Coolify's UI, which also gets TLS.
- **No bring-your-own production database.** Coolify provisions one, or the app
keeps its existing cloud database.
- **GitHub only**, as described under *Limitations*.
- **Portable Postgres has no auth.** Deliberately scoped to the data layer.
- **Switching database mode does not rewrite existing code.** It changes what is
generated from then on; the choice locks once a database is connected.
- **No end-to-end tests.** Unit tests cover the client, tunnel, executor, prompt
and provisioning rules; there is no fake Coolify server yet.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4127?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wwwillchen (commented):
@RyanGroch  this looks promising overall!
if you can create a video, that would be great.

one thing I'm wondering about is how would database backups work? does coolify have an easy way to backup the database to somewhere else? or do we rely on the VPS provider to do backups?

PR #4147, opened 2026-07-29
fix: resume the chat after the database integration Continue

azizmejri1:
`add_integration` parks the local-agent turn on a user-input request. When the user finishes the connector and clicks Continue, the registry arms a follow-up and the chat stream handler is supposed to promote it to `due` on natural completion (`userInputRegistry.streamFinished`), which dispatches "Continue. I have completed the <provider> integration." as the next turn.

The handler gates that on `finishedNaturally`, but the flag was only set on the fall-through return at the end of the try block. The ask, plan and local-agent branches all return early, so the flag stayed false for every agent-mode turn and the `finally` took the `sweepChat` branch instead. The armed request settled as `swept`, no `follow-up-due` was broadcast, and the conversation stopped right after Continue.

Set `finishedNaturally` from `handleLocalAgentStream`'s success result on all three early-return branches. Cancellations and errors still return false, so they keep sweeping as before.

Adds an E2E regression test that drives the whole journey through the packaged app: the agent parks on the request, the user picks a provider in the chat card, finishes the connector in the Configure panel and clicks Continue. It asserts the armed follow-up turn is dispatched into the chat and the card settles into its completed state.

It connects Supabase rather than Neon because `neon:set-app-project` runs `ensureNitroIfVite()`, which installs a Nitro layer into the Vite fixture app; every existing Neon E2E uses the Next.js template for that reason. The regression is provider-independent.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/e2e-tests/fixtures/engine/local-agent/add-integration.ts
+++ b/e2e-tests/fixtures/engine/local-agent/add-integration.ts
@@ -1,0 +1,19 @@
+import type { LocalAgentFixture } from "../../../../testing/fake-llm-server/localAgentTypes";
+
+export const fixture: LocalAgentFixture = {
+  description: "Prompt the user to set up a database integration",
+  turns: [
+    {
+      text: "Let's connect a database first.",
+      toolCalls: [
+        {
+          name: "add_integration",
+          args: {},
+        },
+      ],
+    },
+    {
+      text: "The database integration is connected.",
+    },
+  ],
+};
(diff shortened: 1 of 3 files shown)


wwwillchen:
thanks!

PR #4186, opened 2026-08-03
fix: stop asking for OAuth once an MCP token works

RyanGroch:
*Generated by Claude; reviewed by me*

A server that accepts either OAuth or a personal access token kept showing "This server requires authentication. Enable OAuth and try again." after a working token was saved. Its tools listed and calls through them succeeded, but the alert stayed.

The add dialog has no Headers field, so a token can only be entered on the server's detail page once the server exists. Dyad probes the server in between and gets a 401, which is correct at that moment because there is no token yet. Saving the token did not clear it: the renderer kept that probe result and showed it ahead of the live one. "Enable OAuth & retry" then reported that the server has no OAuth support, and only "Disable OAuth & retry" made the alert go away, because that path probes again.

Build the auth alert from the live tool discovery on each read rather than keeping the probe's answer. Discovery already re-runs whenever a server's config changes, so the alert now ends with the 401 it describes.

The renderer still holds OAuth flow failures, which have nowhere else to live. Its type is narrowed to that one kind so an auth status cannot go back in. The probe is left reporting its 401 as a toast; the retry path reads the discovery its own update triggers.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4186?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/components/plugins/usePluginConnect.test.tsx
+++ b/src/components/plugins/usePluginConnect.test.tsx
@@ -1,0 +1,180 @@
+import { createStore, Provider } from "jotai";
+import { act, renderHook, waitFor } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { ReactNode } from "react";
+import type { McpListToolsResult, McpServer } from "@/ipc/types";
+
+const mocks = vi.hoisted(() => ({
+  listServers: vi.fn(),
+  listTools: vi.fn(),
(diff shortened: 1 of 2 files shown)


wwwillchen:
thanks!

PR #4338, opened 2026-08-21
fix(search-replace): stop writing stray blank lines into edited files

RyanGroch:
Theoretically closes #4286. This might benefit from some more testing/evaluation on search_replace, but I think this fixes the problem that was at the heart of the issue. I'll open a different PR with the extension of the eval suite I used to test this.

The rest of this description was generated by Claude:

search_replace inserted stray blank lines into the files it edited. It hit roughly 1 in 12 edits and accumulated across a session when tested with GPT 5.6 Sol.

The tool joins the model's args with newline delimiters, so a newline that only terminates an arg becomes a trailing empty line in the parsed block. The search side's copy matches nothing and the trimEmptyLines fallback strips it, but nothing stripped the replace side, so the phantom line reached the file. Mirror the fallback's trim onto the replace block, capped at its own edge blanks so a non-empty line is never removed.

Measured on GPT 5.6 Sol across the 16-case search_replace eval suite:

  | metric                                       | count |
  | -------------------------------------------- | ----- |
  | search_replace calls that applied            |    82 |
  | ...where both args ended in a newline        |    18 |
  | ...where that newline matched no blank line  |     7 |
  | ...that wrote a stray blank line             |     7 |

Replaying those calls against the fix: 7 -> 0, byte-identical on the other 75. Matching is untouched — only replaceLines is mutated, after the match index is decided.

Trade-off: a single terminator on new_string meant as "add a blank line", paired with an old_string that mis-terminates a boundary having no blank line, now loses that blank. In 18/18 symmetric calls the replace block ended on the same line as the search block — copied context, not authored blanks — and new_string ending "\n\n" still yields one.

Caveats:
- The asymmetric shape (new_string terminated, old_string not) is unfixed; the fallback never fires for it. Leaving it is justified by the pattern likely being uncommon among models - it was not seen once in 82 calls made by GPT 5.6 Sol.
- applySearchReplace is shared with the <dyad-search-replace> tag path. The new DSL cases cover that path's input shape, but no real tag-path traffic was measured.
- Evaluated with one model, 82 calls.

Also tightens three old_string newline tests from stringContaining to exact output — they passed just as happily on corrupted output, which is why this went unnoticed.

[...]

PR #4345, opened 2026-08-21
test(evals): add newline probe suite and recorded search_replace calls

RyanGroch:
*This is a bit of a slop PR; there's no production code and merging is optional. Description written by Claude.*

Adds the eval harness used to measure how often real models send search_replace arguments in the shape that produced the stray blank lines fixed in #4338, and checks in the recorded calls from that investigation as a regression test.

The probe applies every model search_replace call twice - once with the current serialization, once with a counterfactual that trims a terminating newline - and reports where the two diverge.

recorded_calls/\<model\>/ holds real GPT 5.6 Sol and Luna calls as the operations strings the tool builds from the model's arguments. recorded_calls.spec.ts replays them from fixtures already in the repo and asserts each applied call produces exactly what replacing the first verbatim occurrence of old_string produces. It runs in the normal unit suite, so the evidence behind #4338 stays auditable without an API key and without eval-results/, which is gitignored.

The current processor no longer writes stray blank lines, so the recordings on their own no longer show what they caught. helpers/legacy_search_replace.ts is a frozen copy of the processor from before #4338, and the spec replays both implementations and asserts each model's totals:

  Sol,  before #4338:  9 applied, 7 did not match, 7 stray blank lines
  Sol,  current:      11 applied, 5 did not match, 0 stray blank lines
  Luna, before #4338:  5 applied, 0 did not match, 2 stray blank lines
  Luna, current:       5 applied, 0 did not match, 0 stray blank lines

SEARCH_REPLACE_IMPL=legacy or =current replays just one of them. Sol's applied count goes up because two calls that could not match were blocked by a stray blank line an earlier call had introduced.

Both models produce the defect at different rates - over the full runs these recordings were drawn from, Sol corrupted 7 of 82 applied calls and Luna 2 of 116.

Also adds GPT 5.6 Sol and Luna to the model matrix and EVAL_SKIP_JUDGE=1 for skipping the judge round-trip when only serialization mechanics matter.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4345?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/__tests__/evals/README.md
+++ b/src/__tests__/evals/README.md
@@ -1,8 +1,8 @@
 # Evals
 
-LLM eval suite for tool-use quality. Five suites run the same 16 cases and
-the same three models (Claude Sonnet 4.6, GPT 5.4, Gemini 3 Flash) but with
-different tool sets and system prompts:
+LLM eval suite for tool-use quality. Five suites run the same 16 cases across
+the model matrix (Claude Sonnet 4.6, GPT 5.4, Gemini 3 Flash, GPT 5.6 Sol,
+GPT 5.6 Luna) but with different tool sets and system prompts:
 
 | Suite name               | Tools available                | System prompt                                |
 | ------------------------ | ------------------------------ | -------------------------------------------- |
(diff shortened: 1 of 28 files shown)


wwwillchen (approved):
didn't really review this, but i'm OK with merging this since it's good for future evals.

PR #4366, opened 2026-08-24
fix: make the MCP transport dropdown readable in dark mode

RyanGroch:
The Transport select in the add-plugin dialog used bg-transparent, so the native option list fell back to the browser's white popup while the text inherited the light foreground color. This led to a visual bug where the dropdown items were unreadable.

Give the control the same background and foreground tokens as the other fields in the form.

Before:
DarkModeMCPBefore

After:
DarkModeMCPFixed

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4366?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/components/plugins/AddPluginDialog.tsx
+++ b/src/components/plugins/AddPluginDialog.tsx
@@ -242,7 +242,7 @@ export function AddPluginDialog({
                 data-testid="mcp-transport-select"
                 value={transport}
                 onChange={(e) => setTransport(e.target.value as Transport)}
-                className="w-full h-9 rounded-md border bg-transparent px-3 text-sm"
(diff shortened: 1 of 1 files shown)


wwwillchen (approved):
could you include a screenshot in the PR description? otherwise lgtm.

PR #4378, opened 2026-08-25
Track concurrent chat stream telemetry

azizmejri1:
## Summary

Track when users run chat streams concurrently so product telemetry can measure multi-chat usage without counting duplicate streams for the same chat.

- Emit `chat:concurrent-stream-started` only when admission introduces a second or later distinct chat, while allowing multiple tracked streams for one chat without inflating the count.
- Remove admitted streams during the existing terminal cleanup path so later concurrency counts reflect only live work.
- Bypass non-Pro event sampling for this conversion signal; reviewers should verify that always-on collection is appropriate for the expected event volume.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4378?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/ipc/handlers/chat_stream_handlers.test.ts
+++ b/src/ipc/handlers/chat_stream_handlers.test.ts
@@ -10,6 +10,7 @@ import {
 import { processFullResponseActions } from "@/ipc/processors/response_processor";
 import {
   addTrackedValue,
+  markStreamAdmitted,
   removeDyadTags,
   removeTrackedValue,
   setPartialResponseForStream,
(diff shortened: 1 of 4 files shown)


wwwillchen:
thanks

PR #2997, opened 2026-03-12
fix: hide stale remote-only branches from local delete actions

wwwillchen:
## Summary
- prune remote-tracking refs before listing GitHub branches when auth is available
- track local vs remote-only branch entries in the branch manager instead of flattening them
- disable rename/delete for remote-only branches and add coverage for branch list classification

## Test plan
- npm run fmt
- npm run lint:fix
- npm run ts
- npm test -- --run src/components/GithubBranchManager.test.ts

---

<a href="https://app.devin.ai/review/dyad-sh/dyad/pull/2997" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #4424, opened 2026-08-30
Prune deleted remote branches when refreshing the branch list

### LINKED ISSUE #4171: [bug] Deleted Github Branches stay in dyad locally
After deleting a github branch, it continues to show in Dyad Branches. The branch is deleted in the remote, but stays locally. 
when I try to delete if from dyad - it says it has to be deleted from remote (which it already is)

hdimer:
Fixes #4171.

## The bug

Delete a branch on GitHub (web UI, another machine, a merged PR's auto-delete) and it stays in dyad's branch list forever. Hitting Delete on it then fails with the exact opposite of the truth:

> Branch 'foo' only exists on the remote. To delete it, please delete the branch on GitHub directly.

It has already been deleted on GitHub. There is no way out of that state from inside dyad.

## Two things are wrong

**1. Nothing ever prunes.** `gitFetch()` runs `git fetch <remote>` with no `--prune`, so `refs/remotes/origin/*` keeps entries for branches that are gone. `gitListRemoteBranches()` reads exactly those refs (`git branch -r --list`), and `handleDeleteBranch()` trusts the result, so a stale ref becomes a phantom "remote-only" branch.

**2. "Refresh branches" never talks to the remote.** It calls `refetch()` on the React Query inventory, which re-reads the same stale local refs. So the one control named for this job is a no-op against it.

There is already a `fetch` operation wired end to end for this: it is in the `GithubOperation` union, the transport allowlist, `transition.ts`'s `completionCommands` (which emits `invalidate-branches`), and it has a success banner reading "Fetched latest remote branches". Nothing in the renderer ever dispatches it. `handleFetchFromGithub` is unreachable from the UI today.

## Fix

Two small changes:

- Thread an opt-in `prune` through `gitFetch` and set it on `handleFetchFromGithub`.
- Point "Refresh branches" at the existing `fetch` operation instead of the local `refetch()`.

`invalidate-branches` already refreshes the inventory on completion, so the list updates on its own. The button becomes what its label and the existing banner already claim it is.

Two notes on the judgment calls, since both are yours to make:

- Refresh now costs a network round-trip. That is the point, but it is a behaviour change; the button already sits behind the same `canMutateBranches` idle gate as the rest of that menu, and `GithubBranchManager` only renders once a repo is connected, so the handler's auth and link preconditions cannot fire from here.
- Scope stops at the fetch primitive. Pruning inside `handleListRemoteBranches` instead would put a fetch on every branch-list render and need throttling. The other two `gitFetch` call sites are untouched: fetch-before-rebase does not read the branch list, and connect-repo's fetch does (`prepareLocalBranch` decides whether to create a tracking branch off `origin/<branch>`), so a stale ref there can branch off dead state. That looked like a separate pre-existing bug, so I left it rather than widen this.

You had #2997 open on this in March with a wider scope (prune plus a local/remote badge overhaul with a throttle race) and closed it unmerged. This is a narrower cut. Happy to close it if you would rather take a different one.

## Tests

- `git_utils.test.ts`: real bare remote plus a clone; a branch is deleted directly on the remote, then it asserts plain `gitFetch()` still lists `origin/feature/gone` (pins the bug) and `gitFetch({ prune: true })` drops it while keeping `origin/main`.
- `GithubBranchManager.test.tsx`: pins that "Refresh branches" dispatches the fetch op and no longer re-reads local refs.
- `git_branch_handlers.test.ts`: pins that the Fetch handler passes `prune: true`.

All three verified failing before the change.

`npx oxfmt --check`, `npx oxlint` and `npx tsgo -p tsconfig.app.json --noEmit` are clean.

I used an AI assistant to help investigate and write this; I reproduced the bug and reviewed and tested the change myself.

<a href="https://cubic.dev/pr/dyad-sh/dyad/pull/4424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---
Disclosure: this change was prepared with AI assistance (Claude Code). The repro and tests described above were run before it was opened.

### DIFF
--- a/src/components/GithubBranchManager.test.tsx
+++ b/src/components/GithubBranchManager.test.tsx
@@ -121,6 +121,19 @@ describe("GithubBranchManager machine projection", () => {
     ).toBe(true);
   });
 
+  it("asks the remote to refresh branches instead of re-reading local refs", () => {
+    render(<GithubBranchManager appId={1} />);
+
+    fireEvent.click(screen.getByTestId("branch-actions-menu-trigger"));
+    fireEvent.click(screen.getByTestId("refresh-branches-button"));
+
+    expect(mocks.send).toHaveBeenCalledWith({
+      type: "OP_REQUESTED",
+      op: { type: "fetch" },
+    });
+    expect(mocks.inventory.refetch).not.toHaveBeenCalled();
+  });
+
   it("preserves create input on failure and closes only after success", async () => {
     const view = render(<GithubBranchManager appId={1} />);
 
--- a/src/components/GithubBranchManager.tsx
+++ b/src/components/GithubBranchManager.tsx
@@ -68,7 +68,7 @@ interface BranchManagerProps {
 }
 
 export function GithubBranchManager({ appId }: BranchManagerProps) {
-  const { data, isFetching, refetch } = useGithubBranchInventory(appId);
+  const { data, isFetching } = useGithubBranchInventory(appId);
   const { projection, send } = useGithubOps(appId);
   const {
     capabilities: {
@@ -236,7 +236,9 @@ export function GithubBranchManager({ appId }: BranchManagerProps) {
               Create new branch
             </DropdownMenuItem>
             <DropdownMenuItem
-              onClick={() => void refetch()}
+              onClick={() =>
+                send({ type: "OP_REQUESTED", op: { type: "fetch" } })
+              }
               disabled={!canMutateBranches || isFetching}
               data-testid="refresh-branches-button"
             >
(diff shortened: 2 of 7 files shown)


hdimer:

wwwillchen:
thanks!

